PR #12, opened 2015-01-11
Adding options for showing and hiding the image planes

oscarlorentzon:
Adding options to be able to show and hide the image planes in both walk and orbit mode.

The imagePlaneOpacity property of options is only set to 1 in walk mode since it does not add any value when showing the image planes in orbit mode.

Setting moving mode to walk explicitly before calling the setImagePlaneCamera function in correct circumstances instead of inside function because the method is now also called in orbit mode when a new camera is selected. Creating method for navigating to shot for the same reason.
.

### DIFF
--- a/viewer/reconstruction.html
+++ b/viewer/reconstruction.html
@@ -182,6 +182,7 @@
                 cameraSize: 0,
                 pointSize: 0,
                 showThumbnail: false,
+                showImagePlane: true,
                 drawGrid: false,
                 drawGPS: false
             };
@@ -191,6 +192,7 @@
                 pointSize: 0.7,
                 imagePlaneSize: 50,
                 showThumbnail: true,
+                showImagePlane: false,
                 drawGrid: true,
                 drawGPS: false,
                 animationSpeed: 0.1,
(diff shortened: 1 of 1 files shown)


jesolem:
Awesome @oscarlorentzon

paulinus:
That's great! I love seeing the point cloud when walking. I'll enable depth test on flying mode.

oscarlorentzon:
Hi Pau. Looked at the commits you made with the depthWrite change and the hiding of the imagePlaneOld from the orbit mode view. Although it is nice to have both planes visible for testing purposes i can see advantages of simplifying the orbit mode view to only show the plane for the currently selected camera. 

When i tested with just the first commit (depthWrite) i noticed that the value was not set to true  until after going to walk mode and then back to orbit because the startup values for depthWrite in both materials are still set to false in init(). Therefore imagePlane could still be seen through imagePlaneOld initially. Just wanted to check if this was intended or if I should change the initial value to depthWrite=true to make it correspond to the behaviour in setMovingMode? 

Agree on the point cloud...

paulinus:
yep, good catch! It should be true on init since we start with orbit mode.

PR #14, opened 2015-02-04
Journey

oscarlorentzon:
I really like [this example journey](http://mapillary.github.io/mapillary_examples/routes-malmo_lund.html) and I think it would be really nice to be able to sit back an relax and go on a journey from point A to point B in 3D based on the reconstruction data from multiple sequences. Therefor I have made an attempt with an initial implementation for journeys. 

### Navigation graph

A directed graph is created in python from all the reconstructions for a data set. The edges for each image node are calculated according to the rules for movement defined in the validMoves function in reconstruction.html and the distance is used as weight. The navigation graph is saved to a json file.

### Viewer

The viewer implementation consists of two parts: journey.js and additions to reconstruction.html. The implementation is modular to avoid affecting the current implementation in reconstruction.html.

#### journey.js

journey.js contains three classes: 

* Dijkstra - Calculates the shortest path from the start node to the destination node using Dijkstra's algoritm. 
* Journey - Contains the logic for running a journey. It can be started and stopped and the interval between navigations can be updated. It uses an instance of Dijkstra to calculate the shortest path.
* JourneyWrapper - Uses globally defined functions to create actions used for navigation, UI manipulation and more that are provided to Journey in actions. JourneyWrapper contains the high level API used from reconstruction.html. 

An instance of the journeyWrapper class is created at the bottom of journey.js. 

#### reconstruction.html

reconstruction.html uses the journeyWrapper instance by invoking functions that are run if the journeyWrapper have been defined. The javascript code in reconstruction.html runs even if journey.js is not included in the scripts of the page.

#### Implementation details
 
To promote forward and lateral steps a penalty is added to U-turns and backward steps before calculating the shortest path with Dijkstra's algoritm. If no other directions are available, backward steps or U-turns will still be used to be able to reach the destination node. In special cases this can of course lead to pirouettes, but a pirouette once in a while can not possibly be a bad thing...

## Functionality

To run a journey the following parameters should be specified in the url (which currently can be quite lengthy):

1. nav (required) - The path to the navigation graph.
2. dest (required) - The name of the preferred destination image node.
3. img (optional, reused) - The starting image node.

Both nav and dest need to be supplied for the JourneyWrapper to be initialized and the journey UI elements and functionality to be available. If the img parameter is also supplied the journey will start immediately from that image node, otherwise the UI will load in flying mode as usual and the journey can be started from any node in walk mode.

When the JourneyWrapper is initialized a start/stop button will appear in the navigation area in walk mode. The short command for starting and stopping a journey is the S key. The journey is also stopped by using any of the navigation buttons or commands, making it possible to stop and walk around manually in an interesting area.

The journey can be started or restarted from any node (if a path exist to the destination). The currently selected image node is used as start node and a new shortest path from the current position to the destination node is calculated. 

It is possible to increase or decrease the speed before or during a journey using the animation speed setting in the options. Longer distances between nodes result in longer time between navigations.

## Demo

I downloaded the MINC reconstruction data from the Tech sneak peak blog post to be able to try it out on a large dataset. Below are some example journeys. As you know they take a while to load and they will be blank while loading. 

Note that the old reconstruction.html file from the Tech sneak peak is used because the image coordinates have been changed since then. Therefor some options are missing. For the point cloud journey it is possible to change the speed during the journey since it uses a newer reconstruction.html (where I have removed the show image plane option). 

* [Journey 1](https://dl.dropboxusercontent.com/u/40134832/opensfm/viewer_old/reconstruction_old.html#file=/u/40134832/opensfm/data/minc/reconstruction.json&nav=/u/40134832/opensfm/data/minc/directed_graph.json&res=640&img=WTdbpEsrfJ604ZXZBmQ6ow.jpg&dest=J3IMxpbELquKP3z7EMCgkQ.jpg)
* [Journey 2](https://dl.dropboxusercontent.com/u/40134832/opensfm/viewer_old/reconstruction_old.html#file=/u/40134832/opensfm/data/minc/reconstruction.json&nav=/u/40134832/opensfm/data/minc/directed_graph.json&res=640&img=uPH206tXTOQbs8URgasTig.jpg&dest=HhORroeOUKes-knj1aku2A.jpg)
* [Journey 3](https://dl.dropboxusercontent.com/u/40134832/opensfm/viewer_old/reconstruction_old.html#file=/u/40134832/opensfm/data/minc/reconstruction.json&nav=/u/40134832/opensfm/data/minc/directed_graph.json&res=640&img=aCNbeZpg4JW2wz3hhaMx5w.jpg&dest=-iTXbpAP_eY1mWAlyRBVcQ.jpg)
* [Point cloud journey](https://dl.dropboxusercontent.com/u/40134832/opensfm/viewer/reconstruction.html#file=/u/40134832/opensfm/data/minc/reconstruction.json&nav=/u/40134832/opensfm/data/minc/directed_graph.json&res=640&img=WTdbpEsrfJ604ZXZBmQ6ow.jpg&dest=QHuO5Aqt3VmbWYDRMN7vZg.jpg)

## End note

A whole lot of things can be improved and extended but I think this is a good state for a PR.

### DIFF
--- a/bin/clean
+++ b/bin/clean
@@ -15,3 +15,4 @@ mv -vf $1/hahog $trash
 mv -vf $1/camera_models.json $trash
 mv -vf $1/reference_lla.json $trash
 mv -vf $1/profile.log $trash
+mv -vf $1/navigation_graph.json $trash
--- a/bin/directed_graph
+++ b/bin/directed_graph
@@ -36,12 +36,29 @@ def vector2_angle(x1, y1, x2, y2):
     else:
         return angle
 
-def angle_diff(v, angle):
-    diff = np.abs(v - angle)
+def angle_diff(a1, a2):
+    diff = np.abs(a1 - a2)
 
-    # Make sure the angle difference is less than Pi
+    # Ensure that the angle difference is less than Pi
     diff = np.min(np.vstack((diff, 2*np.pi - diff)), 0)
     return diff
+
+def get_direction_turns_and_motion_angles(min_distance, max_distance, distances, position, direction, positions, directions):
+
+    # Order positions within interval according to shortest distance
+    position_indices = np.where((min_distance < distances) & (distances < max_distance))[0]
+    ordered_indices = position_indices[np.argsort(distances[position_indices])]
+
+    motions = positions[ordered_indices] - position
+    motion_angles = np.array(
+        [vector2_angle(direction[0], direction[1], motion[0], motion[1]) for motion in motions],
+        float)
+    direction_turns = np.array(
+        [vector2_angle(direction[0], direction[1], other_direction[0], other_direction[1])
+         for other_direction in directions[ordered_indices]],
+        float)
+
+    return ordered_indices, direction_turns, motion_angles
 
 def create_optical_center_graph(reconstructions):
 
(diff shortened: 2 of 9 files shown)


ybkuang:
@oscarlorentzon This is super cool!! We will take a closer look tomorrow. 
In the future, it will be awesome that if we also consider the view angle changes in the edge weights, so that the path is smooth (love that you already handling U-turn/backward moving) ;)

paulinus:
Amazing PR!  i've spent some minutes staring at the journeys. So hypnotic ;-)

paulinus:
If we decoupled the image shown in imagePlane and the target of the OrbitControls, it would be possible to smooth the path for the virtual camera. That would produce a smooth motion to destination where the texture of the closest images is shown during the trip.

oscarlorentzon:
Glad you liked it!

Smoother motion would definitely improve the experience. I'll look into it. Thanks for the suggestions!

PR #15, opened 2015-02-10
Updating ceres install instructions.

oscarlorentzon:
I recieved the following error when building the csfm library in Ubuntu:

usr/local/lib/libceres.a(loss_function.cc.o): relocation R_X86_64_32S against `_ZTVN5ceres12ComposedLossE' can not be used when making a shared object; recompile with -fPIC

/usr/local/lib/libceres.a: error adding symbols: Bad value

SuiteSparse needed to be installed according to the shared library instructions in the [Ceres docs](http://ceres-solver.org/building.html) and Ceres needed to be compiled with the -fPIC option. I updated the readme accordingly.

The Ceres Ubuntu installation procedure has been simplified so I removed some outdated information.

I also added correct indentation for code blocks in list items.


paulinus:
The -fPIC option is needed because we are now using ceres from a python extension. Things are still quite shaky since we are trying new things. Thanks for fixing the doc!

PR #17, opened 2015-02-27
PMVS export

ktuite:
I've never made a pull request before so I don't know what I'm doing!

Here's a script to generate output for PMVS. There's a markdown readme to go with it.


ybkuang:
Awesome @ktuite! 

Re: one reason that we made __image_file() private is to ensure DataSet is flexible to use with images locally or in S3. The same applies for the __object_file() methods.

PR #18, opened 2015-03-01
Constant speed journey

oscarlorentzon:
I have made an attempt at enhancing the journey experience by implementing a journey class that moves with constant speed.

## Implementation

A number of changes have been made to journey.js. The following classes have been added or changed. 

* LinearCurve - Inheriting from Three.Curve and is used to retrieve the position in the path at specific points in time.
* JourneyBase - Base class with functions used by inheriting classes such as calculating the shortest path. Defines the empty virtual methods start and stop which should be overridden by classes extending JourneyBase.
* Journey - Extends JourneyBase, otherwise the same functionality as before.
* SmoothJourney - Extends JourneyBase. Continously calculates the current position and viewing direction based on a linear curve created from the nodes in the shortest path when started.
* JourneyWrapper - Improved interval calculation. Functions for adding and removing show path option in menu. Functions for showing and hiding the path to the destination from the selected camera.

## Functionality

The following functionality is available:

* Show path option - If the journey wrapper is initialized a show path option is shown in orbit mode. When checked the path to the destination for the currently selected is drawn. The drawn path and the menu option are removed in walk mode.
* Speed changes - It is possible to increase or decrease the animation speed during the journey with instant feedback.
* Pause - If the animation speed is decreased to zero the journey is "paused" until the animation speed is increased again.

The following url parameter can be specified:

* jou (optional) - If the value is "basic" the old Journey functionality will be available instead of the SmoothJourney.

### Remarks
I have choosen to use the OrbitControls camera for navigation. Because of the behavior of Orbitcontrols the camera will not follow the linear curve exactly. In fact this leads to smoother movement. For more camera control an option would have been to create a dedicated journey camera that can follow the path more accurately. I opted against this for now because of the syncing needed to avoid jumps in space when switching rendering camera.

The preloading in the render method caused tens of images to be downloaded every second leading to black image planes after a while when running journeys. Therefore the valid move preload is disabled during journeys. The journey images are downloaded in small chunks when passing nodes until a buffer of 15 images ahead of the current node is reached. From there only one image at a time is loaded when passing nodes.

To promote forward motion even more than before a small penalty has now also been added to left and right steps and turns.

## Demo

Some demo journeys (with some loading time):

* [Journey 1](https://dl.dropboxusercontent.com/u/40134832/opensfm/s_viewer/reconstruction_old.html#file=/u/40134832/opensfm/data/minc/reconstruction.json&img=aCNbeZpg4JW2wz3hhaMx5w.jpg&nav=/u/40134832/opensfm/data/minc/directed_graph.json&res=640&dest=ReDDL2TthhpguxVwEvgkug.jpg)
* [Journey 2](https://dl.dropboxusercontent.com/u/40134832/opensfm/s_viewer/reconstruction_old.html#file=/u/40134832/opensfm/data/minc/reconstruction.json&img=ikDJrwKmyWCEBA1UX5UMMQ.jpg&nav=/u/40134832/opensfm/data/minc/directed_graph.json&res=640&dest=HhORroeOUKes-knj1aku2A.jpg)
* [Journey 3](https://dl.dropboxusercontent.com/u/40134832/opensfm/s_viewer/reconstruction_old.html#file=/u/40134832/opensfm/data/minc/reconstruction.json&img=df6OKNk1OLqhyxM755UG1w.jpg&nav=/u/40134832/opensfm/data/minc/directed_graph.json&res=640&dest=Es5q50smW0E3dqIfvnD5WA.jpg)
* [Point cloud journey](https://dl.dropboxusercontent.com/u/40134832/opensfm/n_viewer/reconstruction.html#file=/u/40134832/opensfm/data/minc/reconstruction.json&img=HhORroeOUKes-knj1aku2A.jpg&nav=/u/40134832/opensfm/data/minc/directed_graph.json&res=640&dest=ReDDL2TthhpguxVwEvgkug.jpg)

## Further improvments

While being a step forward a lot of things can still improve such as handling vertical angle and amplitude changes as well as smoother direction changes and movement.

### DIFF
--- a/viewer/js/OrbitControls.js
+++ b/viewer/js/OrbitControls.js
@@ -240,6 +240,7 @@ THREE.OrbitControls = function ( object, domElement ) {
 
 		this.target.lerp(this.animationTarget, this.animationSpeed);
 		this.object.position.lerp(this.animationPosition, this.animationSpeed);
+
 		this.object.lookAt(this.target);
 
 		// update condition is:
@@ -254,6 +255,19 @@ THREE.OrbitControls = function ( object, domElement ) {
 			lastQuaternion.copy (this.object.quaternion );
 		}
 	};
+
+	this.adjustAnimationToObject = function () {
+	    this.animationPosition.copy(this.object.position.multiplyScalar(1.01));
+
+	    var direction = new THREE.Vector3(0, 0, -1);
+	    direction.applyQuaternion(object.quaternion).normalize();
+
+	    var target = new THREE.Vector3();
+	    target.addVectors(this.object.position, direction);
+
+        this.target.copy(this.animationTarget);
+		this.animationTarget.copy(target);
+	}
 
 	this.goto_shot = function(cam, shot) {
 		this.animationPosition.copy(opticalCenter(shot));
--- a/viewer/js/OrbitControls.js
+++ b/viewer/js/OrbitControls.js
@@ -240,7 +240,6 @@ THREE.OrbitControls = function ( object, domElement ) {
 
 		this.target.lerp(this.animationTarget, this.animationSpeed);
 		this.object.position.lerp(this.animationPosition, this.animationSpeed);
-
 		this.object.lookAt(this.target);
 
 		// update condition is:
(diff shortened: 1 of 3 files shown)


paulinus:
This is totally awesome @oscarlorentzon !!  So smooth!

oscarlorentzon:
Thanks!

PR #19, opened 2015-03-04
Titles in plots and showing all figures.

oscarlorentzon:
* Adding titles with image names and number of features/matches to plots. 
* Showing all figures for plot_tracks. 
* Fixing typo for feature load in plot_features.


paulinus:
Plotting needed some love, thanks!

PR #20, opened 2015-03-04
Fixing typos in dataset save navigation graph method.

oscarlorentzon:


paulinus:
My bad, did not test after moving. Thanks!

PR #22, opened 2015-04-02
Plot inliers

oscarlorentzon:
To visualize the results for some of the algorithms that includes or excludes feature matches along the pipeline I have created some functions that plot inliers and outliers. 

Two images from a data set needs to specified when running the file but since functions have been extracted a loop can easily be created to plot a number of image combinations.

Because the algorithms are actually run using loaded features and tracks the figures are not created instantly.

The results for two images in the Berlin data set can be seen below:

##### Features and symmetric matches as well as inliers and outliers of robust matches for image 01 and 03:

01 jpg_03 jpg_sift_matches

##### Common tracks (corresponding and not corresponding to robust matches) as well as inliers and outliers of OpenCV find homography and CSfM two view reconstruction:

01 jpg_03 jpg_sift_tracks

### DIFF
--- a/bin/clean
+++ b/bin/clean
@@ -16,3 +16,4 @@ mv -vf $1/camera_models.json $trash
 mv -vf $1/reference_lla.json $trash
 mv -vf $1/profile.log $trash
 mv -vf $1/navigation_graph.json $trash
+mv -vf $1/plot_inliers $trash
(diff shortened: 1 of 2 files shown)


jesolem:
This is awesome @oscarlorentzon !!

oscarlorentzon:
Thanks!

ybkuang:
Great work @oscarlorentzon. This is very useful for visualizing/debugging!

oscarlorentzon:
Thanks again!

niranjantdesai:
Can we include this in the documentation? I almost missed this and was going to write my own visualization functions.

Issue #13, opened 2015-02-01
Segfault in "from opensfm import features"


mtreiber:
I've tried to run the data/berlin sample set using run_all but python is segfaulting.  I've narrowed down the segfault to the bin/detect_features script and the "from opensfm import features" line.  Has anyone run into this before?  I've tried recompiling boost, boost-python, and python from brew but it hasn't seemed to help.  Thanks..

Process:               Python [6959]
Path:                  /usr/local/Cellar/python/2.7.9/Frameworks/Python.framework/Versions/2.7/Resources/Python.app/Contents/MacOS/Python
Identifier:            Python
Version:               2.7.9 (2.7.9)
Code Type:             X86-64 (Native)
Parent Process:        bash [6948]
Responsible:           Terminal [72136]
User ID:               501

Date/Time:             2015-02-01 13:38:40.212 -0500
OS Version:            Mac OS X 10.10.2 (14C81h)
Report Version:        11
Anonymous UUID:        452BF5B4-A444-1CDB-099A-7EE699353D69

Sleep/Wake UUID:       1C147409-D276-4DB7-85BC-A0E8E8801C75

Time Awake Since Boot: 130000 seconds
Time Since Wake:       5800 seconds

Crashed Thread:        0  Dispatch queue: com.apple.main-thread

Exception Type:        EXC_BAD_ACCESS (SIGSEGV)
Exception Codes:       KERN_INVALID_ADDRESS at 0x0000000000000000

VM Regions Near 0:
--> 
    __TEXT                 0000000101e84000-0000000101e86000 [    8K] r-x/rwx SM=COW  /usr/local/Cellar/python/2.7.9/Frameworks/Python.framework/Versions/2.7/Resources/Python.app/Contents/MacOS/Python

Thread 0 Crashed:: Dispatch queue: com.apple.main-thread
0   ???                           	000000000000000000 0 + 0
1   org.python.python             	0x00000001043310dd PyEval_GetGlobals + 23
2   org.python.python             	0x000000010434062b PyImport_Import + 137
3   org.python.python             	0x000000010433ed27 PyImport_ImportModule + 31
4   csfm.so                       	0x00000001041d37b9 init_module_csfm() + 57
5   libboost_python-mt.dylib      	0x0000000104250fd1 boost::python::handle_exception_impl(boost::function0<void>) + 81
6   libboost_python-mt.dylib      	0x0000000104252016 boost::python::detail::init_module(char const*, void (*)()) + 118
[...]

PR #24, opened 2015-04-14
Works with Homebrewed Python in OS X

davll:
Both homebrewed python 2.7 and system bundled Python.framework work fine. :)
(#13 is resolved)


paulinus:
Thanks for solving this @davll !

PR #25, opened 2015-04-20
Journey improvements

oscarlorentzon:
# Journey improvements

I have made the following journey changes to improve the movement:

#### Graph calculation

* Preferred edge distance for navigation graph. Changing image planes too fast makes the movement a little bit hacky. Therefore a config value for preferred distance has been added. Default is set to six meters/units.
* Handling vertical changes in viewing direction and motion that are too large.
* Handling general rotation to avoid to much rotation around camera Z-axis (e.g. images turned upside down).
* Weighting distance, general rotation and motion when calculating the best edges to promote straight movement.
* Config values for navigation thresholds.

#### Client

* To avoid unstable behaviour at the destination node because of skipped nodes the preferred distance graph is merged with the original shortest distance graph on the client. The shortest distance edges are used for nodes within a certain edge distance from the destination.
* Option for movement curve type. Default is movement and viewing direction along spline curves.
* Increased speed in the middle of edges for forward and backward movement through customized speed function.
* Linear image plane transitions (linearish because of the speed function) instead of exponential since the movement is linear.
* Lower maximum speed in turns to avoid very fast turns.

### Demo

Check it out here:

* [Journey 1](https://dl.dropboxusercontent.com/u/40134832/opensfm/i_viewer/reconstruction_old.html#file=/u/40134832/opensfm/data/minc/reconstruction.json&img=aCNbeZpg4JW2wz3hhaMx5w.jpg&nav=/u/40134832/opensfm/data/minc/navigation_graph.json&res=640&dest=ReDDL2TthhpguxVwEvgkug.jpg)
* [Journey 2](https://dl.dropboxusercontent.com/u/40134832/opensfm/i_viewer/reconstruction_old.html#file=/u/40134832/opensfm/data/minc/reconstruction.json&img=uPH206tXTOQbs8URgasTig.jpg&nav=/u/40134832/opensfm/data/minc/navigation_graph.json&res=640&dest=HhORroeOUKes-knj1aku2A.jpg)

### DIFF
--- a/bin/navigation_graph
+++ b/bin/navigation_graph
@@ -155,21 +155,22 @@ def create_optical_center_graph(reconstructions, config):
     step_max_distance = config.get('nav_step_max_distance', 20)
     turn_max_distance = config.get('nav_turn_max_distance', 15)
 
-    max_threshold = np.empty(4)
-    max_threshold.fill(90)
     thresholds = np.radians(np.minimum([config.get('nav_step_forward_view_threshold', 15),
                                         config.get('nav_step_view_threshold', 30),
                                         config.get('nav_step_drift_threshold', 36),
-                                        config.get('nav_turn_view_threshold', 40)],
-                            max_threshold))
+                                        config.get('nav_turn_view_threshold', 40),
+                                        config.get('nav_vertical_threshold', 20),
+                                        config.get('nav_rotation_threshold', 30)],
+                                       [45, 45, 90, 45, 90, 45]))
 
     step_forward_viewing_threshold = thresholds[0]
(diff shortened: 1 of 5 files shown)


jesolem:
Demo looks great @oscarlorentzon !

paulinus:
That is awesomely smooth @oscarlorentzon !

We are making a lot of changes recently that break backwards compatibility of datasets. It does not seem to affect the navigation graph. Merging now!

oscarlorentzon:
Thanks!

The navigation graph uses only the rotation and translation of the shots in the saved reconstruction so the commited dataset changes shouldn't be a problem.

PR #33, opened 2015-05-18
Plot inlier improvements and reprojections

oscarlorentzon:
I have added some figures to the plot inliers script. What you get now if specifying the options to plot everything is the following (Berlin dataset with images 02.jpg and 03.jpg):

##### Features and matches

Same as before with some additional subtitle information like threshold in pixels for each image.

01 jpg_03 jpg_sift_matches

##### Common tracks

Same as before with an added plot at the bottom displaying the result of a bootstrapped reconstruction using the two specified images.

01 jpg_03 jpg_sift_tracks

##### Common reconstructed points

Shows common tracks that are included in the final reconstruction (which is loaded from file). The second subplot shows reprojections for reconstructed 3D points and triangulation reprojections or triangulation failures for rejected tracks. The rejected tracks are triangulated using the final reconstruction cameras. 

01 jpg_03 jpg_sift_reconstruction

The re-triangulation of the rejected tracks can only give an indication of the reason why the tracks were excluded along the pipeline since the camera matrices may change in each bundle adjustment iteration. The re-triangulation is made with a reprojection threshold of 1 so that tracks are not excluded because of the reprojection error.

##### Single image reprojections

Shows all reconstructed tracks for an image reprojected onto the image. Rejected tracks re-triangulated and shown in different colors depending on success or failure.

01 jpg_sift_complete_rec

03 jpg_sift_complete_rec

##### Reprojection error distribution

Shows the reprojection error distribution along with some of its properties like mean and max. Also determines the correlation between the distance from the observations to the principal point and their reprojection error norm. 

01 jpg_sift_error_dist

03 jpg_sift_error_dist

### DIFF
--- a/bin/plot_inliers
+++ b/bin/plot_inliers
@@ -548,8 +548,8 @@ def create_reconstruction_matches_figure(im1, im2, data, save_figs=False):
 
     title = 'Common tracks (loaded): {0}. Reprojected points (calculated): {1}.\n'\
         .format(tracks.shape[0], rec_tracks.shape[0]) + \
-            'Rejected tracks: {0}. Re-triangulated: {1}. Min ray angle ({2} deg) removals: {3}.'\
-        .format(non_rec_tracks.shape[0], succeeded.sum(), min_ray_angle, (~succeeded).sum())
+            'Rejected tracks: {0}. Re-triangulated: {1}. Failed: {2} (behind a camera or min ray angle: {3} deg).'\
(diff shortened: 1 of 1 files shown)


paulinus:
That is awesome again @oscarlorentzon, thanks! Very useful for debugging reconstructions. There's room for improvement in the way we match and bootstrap the reconstruction and plots like this will definitely help.

PR #34, opened 2015-05-18
Plot track

oscarlorentzon:
This PR adds a script that can be used to investigate the result for a particular track for a reconstruction. It plots the epipolar geometry for all combinations of images that are connected to a track. It also displays a 3D figure showing camera positions, principal rays and track feature point rays. 

The following is the output for track 3 of a reconstruction for the Berlin dataset. All three images are connected to the track and their epipoles, epipolar curves, observations and reprojections for the original distorted images are shown. The subtitles show information about the reconstruction distortion parameters as well as distance to epipolar curves and reprojection error.

track_3_sift_epipolar_dist_1

The three cameras are then shown in a 3D plot including image planes, principal points, principal rays, observations, track rays and the reconstructed 3D point.

track_3_ray

There is an option to show the epipolar geometry for undistorted images. Then the images are undistorted using the reconstruction distortion parameters and straight epipolar lines are plotted. The following are the results for track 267 for this reconstruction.

Original distorted images:

track_267_sift_epipolar_dist_1

Undistorted images:

track_267_sift_epipolar_undist_1

Track rays:

track_267_ray

For tracks that were excluded from the reconstruction an attempt is made to make a re-triangulation. If it succeeds the 3D point is shown. When the re-triangulation fails the result resembles the following for track 842 (min ray angle in this case).

Epipolar geometry:

track_842_sift_epipolar_undist_1

Track rays without 3D point:

track_842

At last a figure from another data set showing a track with a larger number of connected images:

kri

### DIFF
--- a/opensfm/reconstruction.py
+++ b/opensfm/reconstruction.py
@@ -224,6 +224,24 @@ def back_project(camera, shot, pixel, depth):
     return np.linalg.solve(A, b)
 
 
+def distort(camera, x, y):
+    """ Distorts a point (x, y) using the distortion coefficients of the camera.
+    :param camera: The camera.
+    :param x: The x coordinate.
+    :param y: The y coordinate.
+    :return: The distorted x and y coordinates.
+    """
+    l1 = camera.get('k1', 0.0)
+    l2 = camera.get('k2', 0.0)
+    r2 = x * x + y * y
+    distortion = 1.0 + r2 * (l1 + l2 * r2)
+
+    x_distort = distortion * x
+    y_distort = distortion * y
+
+    return x_distort, y_distort
+
+
 def reproject(camera, shot, point):
     ''' Reproject 3D point onto image plane given a camera
     '''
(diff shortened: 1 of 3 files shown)


opensfm/reconstruction.py
```diff
@@ -224,24 +224,39 @@ def back_project(camera, shot, pixel, depth):
...
+    yp = camera['focal'] * p[1] / p[2]
 
-    l1 = camera.get('k1', 0.0)
-    l2 = camera.get('k2', 0.0)
-    r2 = xp * xp + yp * yp
-    distortion = 1.0 + r2  * (l1 + l2  * r2)
-
-    x_reproject = camera['focal'] * distortion * xp
-    y_reproject = camera['focal'] * distortion * yp
+    x_reproject, y_reproject = distort(camera, xp, yp)
```
paulinus: Is this computing the distortion the same way as before?
It looks like before the radius r used to compute distortion was the "normalized image coordinates" (X/Z, Y/Z).  Now, it is using the "undistorted image coordinates" (focal * X/Z, focal * Y/Z)

oscarlorentzon:
You are right @paulinus , the distortion computation had been changed. 

I had not realized this before but it is obvious now that the order is important because of the nature of the distortion function and the fact that the distortion parameter should be calculated for coordinates measured in units of focal length according to Hartley 7.4.

Thanks for enlightening me!

The latest commits fixes the distort bug in reconstruction.py and also the faulty distortion logic in plot_track which was caused by the same misconception.

If you think it is better I can revert to the old reproject function in reconstruction.py and move the new distort function to plot_track since it is not used anywhere else. I extracted it to avoid duplication. Or maybe it should just be moved to multiview.py since the undistort_points function is there.

The parameter naming also feels a little bit tricky because of the possible mixup between "scale normalized coordinates" and "focal length normalized coordinates". I tried to make it as clear as possible in the distort function docs.

Great that you spotted my mistake!

oscarlorentzon:
Closing without merging, visualizations not needed.

PR #37, opened 2015-07-20
Error message on missing config.yaml

chris-cooper:
Something like this might help new users get up and running as the error otherwise reported isn't very intuitive.


paulinus:
Thanks for noticing and fixing that @chris-cooper !

PR #59, opened 2016-01-07
Fixing up opengv doc

FreakTheMighty:
I made a small correction to the opengv docs so that they can simply be copy pasted into the shell.


paulinus:
thanks @FreakTheMighty !

FreakTheMighty:
No problem, thanks for the sweet library!

PR #64, opened 2016-01-19
Fix AKAZE std::sort comparison functions

faoiseamh:
std::sort comparison functions should return bool (or values convertible to bool) that satisfy strict weak ordering.

As written, these comparisons return -1, 0, or 1 which has undefined behavior resulting in seg faults.


paulinus:
Thanks for fixing this. That is probably the reason matching with AKAZE didn't work as expected. I've introduced the problem when trying to get a fixed amount of features. I look forward to use AKAZE from opencv3 instead of this version.

faoiseamh:
It depends on how many features you have, but I've been working with AKAZE extensively and can offer feedback if you have any specific questions. Looking forward to OpenCV3's implementation as well

PR #75, opened 2016-04-08
Stream JSON to file

chris-cooper:
Streaming JSON to file reduces memory overhead


paulinus:
Much better. We can probably do the same for loading. Thanks @chris-cooper

PR #87, opened 2016-07-28
Fixed loading graphic for viewer

Caboosey:
Fixed loading graphic since currently http://jimpunk.net/Loading/wp-content/uploads/loading1.gif is not working and being pulled via HTTP instead of HTTPS. Uploaded a good archived copy of graphic via archive.org


paulinus:
Cool, thanks!

Do you think it would be possible to put the changes on a single commit?

Caboosey:
@paulinus Whoever has admin rights can do this. https://github.com/blog/2141-squash-your-commits

EderSantana:
@paulinus you can now setup github to auto squash the commits https://github.com/blog/2141-squash-your-commits

paulinus:
@Caboosey ok, i see why the trial and error commits. Thanks for spotting the problem. I went ahead and made the changes on master (https://github.com/mapillary/OpenSfM/commit/38e79e3e429a9666ee01b77b737250608bca1fe9). We were using a random gif from the net, which we might not have the right to include on the repo. So i've added a gif that we own. Thanks again.

@EderSantana thanks for the tip, didn't know github can squash commits.

PR #98, opened 2016-09-13
Add minfilter to textures in viewer

ekelleyv:
This resolves issues of artifacts along the seams of equirectangular textures. 

Was working on a simple reconstruction viewer using reconstruction.html as a guideline and it took me a while working back from the mapillary-js source to figure out where the artifacts were coming from.

Relevant mapillary-js code:
https://github.com/mapillary/mapillary-js/blob/master/src/component/imageplane/ImagePlaneFactory.ts#L135


paulinus:
Awesome find!  I wondered why that seam was there.
thanks!

PR #99, opened 2016-09-16
Add sub-second time to Exif capture time.

andreasg123:
I'm using OpenSfM with images extracted from video at more than 1 fps.  Thus, it is desirable to have sub-second accuracy for the capture time when using `matching_time_neighbors`.

This PR checks the corresponding Exif SubSec tags and adds the value to the timestamp.


paulinus:
Great addition, thanks @andreasg123 !

PR #103, opened 2016-09-26
Write profile data for create_tracks

andreasg123:
When checking which steps of `run_all` make use of multiple CPU cores, I noticed that `create_tracks` is the only command called by `run_all` that doesn't add to `profile.log`.  For my data set, it takes longer than the other fast commands, due to 24 million calls to `UnionFind.union`.


paulinus:
Perfect, thanks @andreasg123 !

PR #104, opened 2016-09-27
Use multiprocessing when finding homography between image pairs

andreasg123:
For a large reconstruction (50 seconds of video at 8 fps with neighbors +/- 15 seconds), it took about 15 minutes to set up the image pairs before the actual reconstruction started.  This PR reduces that time greatly by using multiprocessing on a system with 16 cores, 32 hyper threads (1 process: 859 seconds; 32 processes: 63 seconds).

Unfortunately, the time for calling `common_tracks` for all image pairs, not included in the above, is 50 seconds.  When attempting to include it in the multiprocessing, the time for pickling `graph` dominated and most workers were idle.  This was true even when grouping the tasks by `im1`.  It may be possible to get around this by passing the graph as an initializer for each worker but I didn't try that.

### DIFF
--- a/opensfm/reconstruction.py
+++ b/opensfm/reconstruction.py
@@ -3,13 +3,14 @@
 
 import datetime
 import logging
-from itertools import combinations
+from itertools import combinations, groupby
 
 import numpy as np
 import cv2
 import pyopengv
 import time
 from networkx.algorithms import bipartite
+from multiprocessing import Pool
 
 from opensfm import align
 from opensfm import csfm
@@ -176,19 +177,34 @@ def pairwise_reconstructability(common_tracks, homography_inliers):
 
 def compute_image_pairs(graph, image_graph, config):
     """All matched image pairs sorted by reconstructability."""
-    pairs = []
-    score = []
-    for im1, im2, d in image_graph.edges(data=True):
-        tracks, p1, p2 = matching.common_tracks(graph, im1, im2)
-        if len(tracks) >= 50:
-            H, inliers = cv2.findHomography(
-                p1, p2, cv2.RANSAC, config.get('homography_threshold', 0.004))
-            r = pairwise_reconstructability(len(tracks), inliers.sum())
-            if r > 0:
-                pairs.append((im1, im2))
-                score.append(r)
+    args = _pair_reconstructability_arguments(graph, image_graph, config)
+    processes = config.get('processes', 1)
+    if processes == 1:
+        result = map(_compute_pair_reconstructability, args)
+    else:
+        p = Pool(processes)
+        result = p.map(_compute_pair_reconstructability, args)
+    pairs = [(im1, im2) for im1, im2, r in result if r > 0]
+    score = [r for im1, im2, r in result if r > 0]
     order = np.argsort(-np.array(score))
     return [pairs[o] for o in order]
+
+
+def _pair_reconstructability_arguments(graph, image_graph, config):
+    homography_threshold = config.get('homography_threshold', 0.004)
+    return [(homography_threshold, im1, im2, d, matching.common_tracks(graph, im1, im2)) 
+            for im1, im2, d in image_graph.edges(data=True)]
+
+
+def _compute_pair_reconstructability(args):
+    homography_threshold, im1, im2, d, common_tracks = args
+    tracks, p1, p2 = common_tracks
+    r = 0
+    if len(tracks) >= 50:
+        H, inliers = cv2.findHomography(
+            p1, p2, cv2.RANSAC, homography_threshold)
+        r = pairwise_reconstructability(len(tracks), inliers.sum())
+    return (im1, im2, r)
 
 
 def get_image_metadata(data, image):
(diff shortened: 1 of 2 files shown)


andreasg123:
I tried to encapsulate the graph in a function object but pickling the graph still keeps most workers idle.

    class _PairReconstructability:
        def __init__(self, graph):
            self.graph = graph
        def __call__(self, args):
            homography_threshold, im1, im2, d = args
            tracks, p1, p2 = matching.common_tracks(self.graph, im1, im2)
            r = 0
            if len(tracks) >= 50:
                H, inliers = cv2.findHomography(
                    p1, p2, cv2.RANSAC, homography_threshold)
                r = pairwise_reconstructability(len(tracks), inliers.sum())
            return (im1, im2, r)

I cannot think of another way to make `common_tracks` run in the pool workers without getting slowed down by pickling the graph.

opensfm/reconstruction.py
```diff
@@ -176,21 +177,36 @@ def pairwise_reconstructability(common_tracks, homography_inliers):
...
+        result = p.map(_compute_pair_reconstructability, args)
+    pairs = [(im1, im2) for im1, im2, r in result if r > 0]
+    score = [r for im1, im2, r in result if r > 0]
     order = np.argsort(-np.array(score))
     return [pairs[o] for o in order]
 
 
+def _pair_reconstructability_arguments(graph, image_graph, config):
+    homography_threshold = config.get('homography_threshold', 0.004)
+    return [(homography_threshold, im1, im2, d, matching.common_tracks(graph, im1, im2))
```
paulinus: Would it work to pass `graph` as an argument instead of `matching.common_tracks(graph, im1, im2)` and then compute the common_tracks in `_compute_pair_reconstructability`?

andreasg123: No.  Passing `graph` as an argument means that the graph has to be pickled to be passed to the other processes.  That makes the main process run at 100% CPU, presumably for pickling, while the other processes created by multiprocessing are mostly idle.  With the current approach, all the additional processes run at 100% CPU.  As I wrote in my comment, I also tried to encapsulate the graph in the function being called but that did not improve the situation.

Presumably, this would work with multithreading instead of multiprocessing because the graph wouldn't need to be pickled.  However, while `multiprocessing.dummy` uses threads, it still pickles all parameters.  One would have to implement another version of `Pool.map` with threads.

andreasg123: [Multithreading ](https://docs.python.org/2/library/threading.html)does exist in Python and [multiprocessing.dummy](https://docs.python.org/2/library/multiprocessing.html#module-multiprocessing.dummy) is one way to make use of it, unfortunately still using a queue containing pickled parameters for function calls.  I haven't looked for other threading modules that would maintain a queue for worker threads.

paulinus: Also, i think that the GIL will prevent the threads from running in parallel

andreasg123:
The approach that I suggested for [pickling the graph](https://github.com/mapillary/OpenSfM/pull/106#issuecomment-254818773) still is too slow for my large data set with 28 million total image tracks (instead of 200,000 that I used for my pickle benchmarks).  I'm now ready to give up on the idea to use the graph as a parameter for multiprocessing.

I added another slight improvement that reduces the time from entering `incremental_reconstruction` to the first call to `bootstrap_reconstruction` from 171 seconds to 121 seconds.  Instead of first calling `bipartite.weighted_projected_graph` to determine connected images and later calling `common_tracks` for each pair, I created a function `all_common_tracks` that combines the two steps.  The final savings would be a little larger because I didn't include the subsequent calls to `common_tracks` from `bootstrap_reconstruction` in the comparison.

Due to the use of RANSAC in `two_view_reconstruction`, it is not possible to compare the outputs of the two versions.  However, I sorted the tracks alphabetically in both `common_tracks` and `all_common_tracks` and verified that the same tracks and features are used in `bootstrap_reconstruction`.

This approach could be adapted for `compute_depthmaps` by passing `reconstruction.points` as the second parameter of `all_common_tracks` to just consider those tracks.  `all_common_tracks` could be called with `include_features=False`, greatly speeding it up, because the features aren't used in `find_neighboring_images`.

paulinus:
Thanks for the analysis and solution @andreasg123.  Let me summarize my thinkings on this to make sure i get the situation right.

We need to do some work for every image pair. The work is essentially

1.  find common tracks and
2. estimate an homography

The second step is the slowest and we want to run it in parallel.

We would like to run 1 in parallel too, but that requires passing the graph to the worker processes and that is too expensive on its own.  So we do 1 sequentially in the `all_common_tracks` that you implemented.

It all looks good to me, i'm going to merge and do some minor clean up.  Thanks again!

andreasg123:
@paulinus, your summary is perfect.

bryanibit:
@andreasg123 @paulinus   Could you explain for me  why we need to calculate outlier ratio through homography and make sure the radio is more than 0.3, and then add the pair to reconstruction pairs? Thank you! I am really confused about that.

paulinus:
Reconstruction of the initial pair is ill-defined and fails if the scene is planar or there is no parallax. The homography is used to identify such cases and discard them. If 70% of the matches can be explained by an homography, we discard the pair.

I'm currently working on an alternative initialization for the planar case which is common in drone imagery.

The "no parallax" case correspond to pairs of images taken without translating the camera, like when you take a panorama. In such case, it is not possible to recover 3D information from the correspondences.

bryanibit:
Thank you for your reply. You are understood. The pure rotation and the plane for a pair of views will result in failure in initial reconstruction.

PR #105, opened 2016-10-04
Adds eigen dependency to Readme for OSX

spikeheap:
This was necessary to get OpenGV to install on OSX Sierra for me.


paulinus:
Yep, that is required. Thanks for adding it!

PR #106, opened 2016-10-14
Parallelize depthmap computation

mojodna:
I was also working on this last night but hadn't had a chance to test it due to https://github.com/OpenDroneMap/OpenDroneMap/issues/397.

I'd like to propose this as an alternative to 8f50c909b75eb51a275f4f5b8745a44d263dbe7d, as it keeps the function signatures intact (though shifting `shot` to be the last arg in `find_neighboring_images` so that it can be partially applied). If you like it, I can rebase it.

@paulinus thoughts?

### DIFF
--- a/opensfm/dense.py
+++ b/opensfm/dense.py
@@ -1,5 +1,7 @@
 
+from functools import partial
 import logging
+from multiprocessing import Pool
 
 import cv2
 import numpy as np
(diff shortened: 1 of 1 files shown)


paulinus:
oh, sorry for the clash @mojodna 

I really like the partial trick, and keeping the function signatures. I could attempt a merge but it will be pretty violent i think since i've changed a few other things like removing the return values.
It would be great if you could you rebase, or rather just reapply, the signature trick over the current master.

Also, i still don't know if computing the neighborhood in parallel is worth it, since spawning a new process is pretty expensive (see also #104). For now we can keep that part serial.

thanks!

mojodna:
Gotcha will do. Hopefully tomorrow, otherwise soonish (though it could be next week). No great rush, as there's already a good solution in there.

The other spot I was going to add parallelism was in [`focal_from_exif`](https://github.com/mapillary/OpenSfM/blob/master/bin/focal_from_exif#L23), as that's one of the early spots in ODM that seems like it could be marginally quicker.

paulinus:
cool! go for it!
thanks

andreasg123:
Did you look at CPU utilization with `top` after applying this parallelization?  In #104, I found that anything that passes `graph` to the parallel functions takes too much time for pickling the graph.  The result is that only a few of the spawned processes use 100% CPU because pickling `graph` for the queue takes so long.

If there were a way to partition the graph such that only a small subgraph would need to be passed to each process, this wouldn't be an issue.

paulinus:
Yes, just passing the graph is taking too much time, although still a small percentage of the time it takes to compute the depthmap.
Splitting the graph is possible since we only use the edges around a particular image. I think for this case, not passing the graph at all will be better. We can compute the depth range outside.

andreasg123:
I conducted benchmarks with a reconstruction with 137 images and 200,000 total image tracks.  I don't know if `multiprocessing` uses `pickle` or `cPickle`.  Also, I created a dictionary that only contains the information needed by `matching.common_tracks`.

    {im: {track: {'feature': graph[im][track]['feature']} for track in graph[im]} for im,d in graph.nodes(data=True) if d['bipartite'] == 0}


| Operation | Seconds |
| --- | --- |
| pickle graph | 11.3 |
| cPickle graph | 2.6 |
| create dict | 0.2 |
| cPickle dict | 1.2 |
| un-cPickle dict | 0.4 |

For any parallel task that would take noticeably longer than 0.4 seconds, it would be worthwhile to create the dictionary, pickle it once, pass the string to the parallel function, and unpickle it in the function.

I'll run a benchmark to see if this approach makes sense for #104.

varadgunjal:
If I understand correctly, the update proposed here was implemented within #104 and merged with the master right?

domdomegg:
@paulinus Now #104 is merged, can this be closed?

YanNoun:
Closing, as `dense` is now running as a combination of multithreading + C++ GIL release.

Issue #118, opened 2016-12-08
Image masks


jer5:
In some circumstances it would be useful to have certain parts of each image ignored (for example if some parts of the image are known to contain moving objects which would confuse the reconstruction process). Therefore it would be useful to have the option to supply a binary pixel mask with each image to indicate which parts of the image should be ignored.

The simplest way to supply the pixel mask is probably to have another mask image along side each real image. Alternatively an alpha channel on a single image could be abused to pass the information.

Using a specific colour in the image to indicate the masked area (e.g. setting all areas to be ignored to black) would be less useful as this would lose the image of the moving objects in the original image.

Another use case for this mask capability is where camera lens correction has been applied to the images, resulting in images that have a region around their edges where only some of the pixels contain valid data from the raw image.

paulinus:
That would be a nice feature to add @jer5.  The simplest way to do that would be to remove any feature outside the masks right after feature detection.

We have a very basic mask concept in place, which is a list of rectangles to ignore on every image. The use case for that is to remove static watermarks.

A binary bitmap per image would be better.

The code to modify is around https://github.com/mapillary/OpenSfM/blob/f80abbcb09a0ec6b016b967f291aa09f6567c39f/opensfm/features.py#L71-L86

Do you want to try implementing it?

StrausMG:
Could you clarify why the issue is not closed if #132 is merged?

PR #132, opened 2017-01-24
Allow per image masking

BrookRoberts:
Adding additional feature where you can mask out parts of the image on a frame by frame and pixel by pixel basis, rather than having to specify static rectangles for the whole range.

Masks should have the same name as the corresponding image, and are listed in the same way as images, i.e. you can provide a textfile listing them, or you can just dump them in a folder next to images, called 'masks'.  They should be image files of the same dimension as the original images.  The first channel is scanned, and if it is equal to 0 then features there are kept, and if it is non-zero they aren't. (i.e. for a black and white mask, any white areas of the mask are deemed not useful, e.g. watermarks, moving objects, and so on).

Is there any relevant documentation to update?  I've also not really looked into the tests associated with OpenSFM, so would appreciate some pointers there if appropriate.  I did also for my own testing make something that would output images of the location of features found, before and after mask filtering.  I don't think these should be included, but am happy to put them somewhere if useful.  Was certainly useful for being really sure what was going on.

#118

### DIFF
--- a/opensfm/commands/detect_features.py
+++ b/opensfm/commands/detect_features.py
@@ -41,13 +41,14 @@ def detect(args):
     logger.info('Extracting {} features for image {}'.format(
         data.feature_type().upper(), image))
 
-    mask_path = None
-    mask_name = image.split('.')[0] + 'mask'
+    mask_name = image
     if mask_name in data.masks():
         mask_path = data.mask_files[mask_name]
         logger.info('Found mask {} to apply'.format(
             mask_name
         ))
+    else:
+        mask_path = None
 
     if not data.feature_index_exists(image):
         preemptive_max = data.config.get('preemptive_max', 200)
--- a/opensfm/commands/detect_features.py
+++ b/opensfm/commands/detect_features.py
@@ -41,19 +41,13 @@ def detect(args):
     logger.info('Extracting {} features for image {}'.format(
         data.feature_type().upper(), image))
 
-    mask_name = image
-    if mask_name in data.masks():
-        mask_path = data.mask_files[mask_name]
-        logger.info('Found mask {} to apply'.format(
-            mask_name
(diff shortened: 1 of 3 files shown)


paulinus:
This is really great @BrookRoberts ! Thanks for adding that.

I only have minor comments on it
- We try to avoid opening files from disk outside the `Dataset` class. This is because, we have alternative versions of the `Dataset` class that load files from the network. So instead of loading the masks in `features.py`, it would be better to have a method like `Dataset.mask_as_array(image)` in `dataset.py` 
- Really minor: I don't think there is any standard, but i'm used to consider 0 as the mask out value. White is keep and black is ignore. Would it be ok for you to reverse?

BrookRoberts:
Thanks!

First comment sounds reasonable, I'll change that.
Second comment - yes, I guess that makes slightly more sense.  It wasn't entirely obvious to me which way made more sense and my way fit better with my current outputs!

I'll try and fix this up soonish.

paulinus:
awesome, thanks!

BrookRoberts:
Right, have made a commit that hopefully addresses those issue.

paulinus:
Good fixes @BrookRoberts!  I had a problem when testing with images larger than 2048 since the width and height of the image is then smaller than the size of the mask. I have a fix working. I'll merge this PR and commit the fix after that.

paulinus:
I'm also going to change the name of the mask files so that the mask of `images/im1.jpg` is `images/im1.jpg.png`. It is a bit ugly to add a second extension to a file name, but it makes it very easy to map the image name to the mask file.

This makes it easy to have masks in PNG format regardless of the format of the original images. Otherwise, JPG images need to have JPG masks, which is not a good format for binary images.

I hope that works for you. Thanks for the PR!

Issue #82, opened 2016-06-16
Consecutive images


profdunny:
Is there any function to process consecutive equirectangular images with loop closing? I am interested in large-scale reconstruction using an omnidirectional camera, however, it takes very long time to calculate matches between all frames.

paulinus:
Assuming you have GPS position of the images on the EXIF, you can set the
option `matching_gps_neighbors` to something like 4 to limit every image to
be matched only to the closest 4 neighbors. This uses the GPS position to
determine the neighbors.

Try adding the line
```
matching_gps_neighbors: 4             # Number of images to match selected
by GPS distance. Set to 0 to use no limit
```
to the `config.yaml` of your dataset.

dunny2016:
Thank you for your kind reply.

I don't have GPS position data. I extract sequential images from the omnidirectional video.
Is there any function to match sequential images?

mapninja:
I just want to bump this, because I'm interested in this, too. Perhaps if we wrote "faked" GPS data to the EXIF header?, but it seems like this should be a core setting, especially for screen grabs from video.

jer5:
This would be useful to us too. Images that have come from video have a natural sequence and it would be useful to use this to determine the neighbours instead of GPS position (which is not always available or valid).

I suggest the following three possibilities for passing the sequence number information in.

1) The sequence number could be generated from the exif attribute "Exif.Image.DateTimeOriginal".

2) The sequence number could be taken directly from the exif attribute "Exif.Image.ImageNumber".

3) The sequence number could be embedded within the image file name. E.g. "frame00001.jpg", "frame00002.jpg" etc. in a similar way to that which ffmpeg generates output image file names when given the output name "frame%05d.jpg".

paulinus:
Since the merge above, you can specify the option `matching_order_neighbors: 2` to match only images that are consecutive when sorted by image name.

PR #133, opened 2017-01-24
Image matching on consecutive images

BrookRoberts:
Adds extra configuration option to match on consecutive images ordered by name.  Use case is when you have a sequence of images, e.g. ripped from a video, that are in linear order.  Substantially speeds up the matching step in these cases.

#82


paulinus:
Smart solution! we though that time-based filtering could be used for that use case, but exif is not always available, so great to have an option to use filenames instead.
thanks!

Issue #126, opened 2017-01-02
opensfm.io.json_dump minification


j6k4m8:
Minifying the `reconstruction.meshed.json` file can dramatically reduce the filesize: In one example, my 267MB file shrunk to 63MB. The original file failed to load in the web-viewer: My minified version loaded immediately.

Is there a reason the whitespace is left in (perhaps for human-readability)? If not, I'd be happy to pull-request this change.

> For those who just want to try minifying their files too, you can run:
> ```python
> import json
> with open('reconstruction.meshed.min.json', 'w') as fh:
>     fh.write(
>         json.dumps(
>             json.loads(open('reconstruction.meshed.json', 'r').read())
>         )
>     )
> ```

BrookRoberts:
This problem and solution both affected me too.  Ideally the reconstruction viewer would also warn you in some way that the file was too big rather than hanging.  Certainly I couldn't get a file to load at 486MB but the minified 188MB worked fine.

paulinus:
@j6k4m8 the white space is indeed only for human-readability. Happy to accept that PR and remove it, thanks. A human can use a json formatter if he needs to read it.

PR #144, opened 2017-02-16
Minify reconstruction.meshed.py

BrookRoberts:
reconstruction.meshed.py gets too large to be loaded in the browser viewer for large datasets, and minifying this mitigates this. See #126 

I wasn't sure exactly how much to minify.  Setting indent to 0 rather than 4 gets most of the space saving gain, and you can still open the files easily in text editors, they just look less nice.  Setting indent to None removes new lines as well, and then you can't open the files easily in text editors and it would be very hard to human read.
For Lund I got size of 7.7mb for indent=4, 2.2mb for indent=0, 1.9mb for indent=None.

Since it's nice to have output human readable unless there is a good reason not to, I gave json_dump an argument to minify, and only activate for this one file at the moment.  But happy to change to minify all files/some other subset.

### DIFF
--- a/opensfm/commands/mesh.py
+++ b/opensfm/commands/mesh.py
@@ -31,7 +31,8 @@ class Command:
                     shot.mesh.faces = faces
 
         data.save_reconstruction(reconstructions,
-                                 filename='reconstruction.meshed.json')
+                                 filename='reconstruction.meshed.json',
+                                 minify=True)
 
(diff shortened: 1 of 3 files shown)


paulinus:
That's great @BrookRoberts!  It is good to have the option. We can always switch the default to `True` if we need so.
Thanks, merging.

PR #147, opened 2017-02-27
Update sensor_data.json

merkato:
Added DJI Phantom 3 Standard


paulinus:
adding. Thanks @merkato !

PR #148, opened 2017-02-28
Compilation fixes for non x86 architectures

EmbeddedAndroid:
Two small fixes to allow OpenSFM to compile on non x86 architectures.

I'm trying to upstream all the fixes to allow OpenDroneMap run an ARM processors. For reference:
OpenDroneMap/OpenDroneMap/issues/484


paulinus:
Great @EmbeddedAndroid. I can't test on ARM so trusting you on that one. Thanks!

EmbeddedAndroid:
@paulinus thanks! compiled test on both ARM and x86!

PR #151, opened 2017-03-16
Making OpenSFM install more nicely

sinjax:
`python setup.py install` now makes it so a version of run_all can be ran globally


paulinus:
Awesome! thanks for the fix!  I'll delete the original run_all, and at some point move it to a subcommand of `opensfm` so you can do `opensfm run_all`

PR #154, opened 2017-03-27
Update documentation to use virtualenv

kachayev:
README file now suggests using `virtualenv` to create isolated `python` environment to manage installed dependencies separately from system-level packages.


paulinus:
Thanks for the instructions.  I have some questions

- Running `virtualenv .` mixes virtualenv tools with opensfm tools. Isn't it better to use a different folder?
- Should we skip the line installing `scipy` and add it to `requirements.txt` instead?

kachayev:
@paulinus Sure, that makes sense. 

1. Updated README to use `env` as an environment name, added `env` directory to `.gitignore` file.

2. I was wondering why `scipy` isn't there already, that's why I added it as a separate installation step. But if it was not done intentionally, it's obviously better to keep all dependencies in a single file. There's another question here: all dependencies are mentioned w/o version specification, which leads to a huge potential problem with reproducibility of the build. I attached `requirements.txt` file with concrete versions to avoid any problems in future.

kachayev:
@paulinus Removed version specification for `numpy` and `scipy`, due to problems with Linux builds. It works perfectly fine with `numpy==1.12.1` but fails on TravisCI.

paulinus:
@kachayev I don't know what is the best way to handle versions in requirements.txt.

Using specific versions has the advantage of reproducibility, but it is a bit misleading. OpenSfM does not actually require those specific versions.

For example, I'm using different versions than the ones you added. And you seem to have a different version of numpy than the ones available in the Travis build.

I don't know what is the best practice for this problem. Do you have a reference?

zwn:
https://packaging.python.org/requirements/#requirements-files

Minimal requirements go to setup.py while version pinned go to a requirements file.

paulinus:
Exactly what I was looking for. Thanks @zwn !

I'll then merge this PR and we can add pinning for numpy later.

PR #185, opened 2017-06-08
Match signature by converting id to string

BrookRoberts:
Seemed not to match the signature required - note in the normal bundle this conversion is applied.


paulinus:
perfect, thanks @BrookRoberts !

PR #186, opened 2017-06-08
Local bundle update

BrookRoberts:
Merge master into feature-local bundle, and fix both cases where shot.id needs to be converted to string (last commit)


paulinus:
@BrookRoberts does this merge include the incremental matching of this other PR #178

I'll merge `master` to `feature-local-bundle` independently first and then see what are the remaining differences with `BrookRoberts/feature-local-bundle`.

BrookRoberts:
@paulinus I wasn't intending it to.  I'm currently using local-bundle for somethings, and noticed it was quite a bit off master.

paulinus:
I've merged the branch to master so that it does not go off anymore.
Added a parameter in config to use the local bundle feature so that it is off by default

PR #187, opened 2017-06-23
Convert shot.id to string as required by bundle adjuster

BrookRoberts:
The fix for bundle_single_view appears to be in master but the fix for bundle_local is not - I think it must have just got lost in the merge.


BrookRoberts:
Similar to #185

paulinus:
Probably did not fix the `bundle_local` function before since it was on another branch.

PR #188, opened 2017-06-28
When loading shots, load mesh if it exists

BrookRoberts:
Currently if you use io to load shots, the mesh doesn't get loaded even if it exists.


paulinus:
good catch, thanks!

PR #195, opened 2017-07-06
fix(match_features): allow matching_gps_neighbors to be unset in config

erikreed:
Fixes an exception where using `matching_gps_neighbors: 0` or omitting matching_gps_neighbors in the config prevents feature matching from completing. Frames without GPS info will error out using the default config values.

This PR prevents the following exception:
```
  File "/source/OpenSfM/bin/opensfm", line 34, in <module>
    command.run(args)
  File "/source/OpenSfM/opensfm/commands/match_features.py", line 26, in run
    pairs = match_candidates_from_metadata(images, exifs, data)
  File "/source/OpenSfM/opensfm/commands/match_features.py", line 166, in match_candidates_from_metadata
    gps_neighbors, max_distance)
  File "/source/OpenSfM/opensfm/commands/match_features.py", line 94, in match_candidates_by_distance
    gps['latitude'], gps['longitude'], gps['altitude'],
KeyError: 'latitude'
```

### DIFF
--- a/opensfm/commands/match_features.py
+++ b/opensfm/commands/match_features.py
@@ -154,9 +154,10 @@ def match_candidates_from_metadata(images, exifs, data):
         data.invent_reference_lla()
     reference = data.load_reference_lla()
 
-    if not all(map(has_gps_info, exifs.values())) and gps_neighbors != 0:
-        logger.warn("Not all images have GPS info. "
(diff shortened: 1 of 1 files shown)


paulinus:
The warning was confusing. Thanks for the fix.

PR #196, opened 2017-07-07
Optimize Dockerfile: cache pip install

erikreed:
Currently, changing a single line of code and re-building a docker image ends up re-installing the python dependencies which takes a while. The pip install can be moved to the top of the Dockerfile so this layer is cached and unaffected by changes to unrelated files.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,8 +1,9 @@
 FROM paulinus/opensfm-docker-base
 
-COPY . /source/OpenSfM
+COPY requirements.txt /source/OpenSfM/requirements.txt
+RUN pip install -r /source/OpenSfM/requirements.txt
 
+COPY . /source/OpenSfM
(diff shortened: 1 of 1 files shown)


paulinus:
We put everything on a single docker command because it is supposed to make docker images smaller. We can change it if needed.

Are you usually in the case where you modify the code and run docker build again?

I think there must be a way of modifying the code and getting it used inside the docker container without rebuilding an image. This way you would not need to run `python setup.py build` neither when you just modify python code.

erikreed:
You're quite right -- this was a bit over-optimized for my use case. Although the image size should be the same (sans metadata), unless the build step performs any cleaning of dependencies installed via pip (as the final image size is a sum of the layer diffs, and the last layer here is split into two diffs with no intersection and same end state).

PR #197, opened 2017-07-07
Fix config feature-matching comments

erikreed:
The current comments don't reflect the behavior of:
https://github.com/mapillary/OpenSfM/blob/d96023b41606176d34a988443fca423444f27258/opensfm/commands/match_features.py#L111
```python
if max_neighbors <= 0:
    return set()
```

That is, 0 does not mean no limit; it means bail out of matcher. This PR remedies that.


paulinus:
The casuistic there is a bit messy. Your comments make it more clear.

PR #211, opened 2017-09-03
Update Mac OSX dependencies installation instruction

IanBoyanZhang:
Thanks for providing great tools!  homebrew/science/ceres-solver is deprecated.


domdomegg:
8a4c4e7 should be only in #210, not here too

IanBoyanZhang:
Thanks for pointing that out. Will fix it.

Issue #198, opened 2017-07-09
Matching fails if at least one image has no GPS data


jleistner:
I've installed the latest version of OpenSfM today and now all my data sets don't work anymore. From the console output below you can see that the matching fails but I can't figure out why.
[townhall.zip](https://github.com/mapillary/OpenSfM/files/1133744/townhall.zip)
```
bin/opensfm_run_all data/townhall
2017-07-09 14:36:09,696 Extracting focal lengths for image IMGP3419.jpg
2017-07-09 14:36:09,700 Extracting focal lengths for image IMGP3428.jpg
2017-07-09 14:36:09,702 Extracting focal lengths for image IMGP3420.jpg
2017-07-09 14:36:09,704 Extracting focal lengths for image IMGP3417.jpg
2017-07-09 14:36:09,706 Extracting focal lengths for image IMGP3425.jpg
2017-07-09 14:36:10,079 Extracting ROOT_HAHOG features for image IMGP3419.jpg
2017-07-09 14:36:10,079 Extracting ROOT_HAHOG features for image IMGP3428.jpg
2017-07-09 14:36:10,079 Extracting ROOT_HAHOG features for image IMGP3420.jpg
2017-07-09 14:36:10,080 Extracting ROOT_HAHOG features for image IMGP3417.jpg
2017-07-09 14:36:10,080 Extracting ROOT_HAHOG features for image IMGP3425.jpg
2017-07-09 14:36:11,178 Found 4827 points in 0.954954862595s
2017-07-09 14:36:11,183 Found 4889 points in 1.01493597031s
2017-07-09 14:36:11,407 Found 4888 points in 1.18394899368s
2017-07-09 14:36:11,476 Found 4885 points in 1.27148008347s
2017-07-09 14:36:11,774 Found 4809 points in 1.55314683914s
2017-07-09 14:36:12,325 Matching 0 image pairs
2017-07-09 14:36:12,350 Matching IMGP3428.jpg  -  1 / 5
2017-07-09 14:36:12,350 Matching IMGP3425.jpg  -  2 / 5
2017-07-09 14:36:12,351 Matching IMGP3419.jpg  -  3 / 5
2017-07-09 14:36:12,351 Matching IMGP3417.jpg  -  4 / 5
2017-07-09 14:36:12,351 Matching IMGP3420.jpg  -  5 / 5
2017-07-09 14:36:12,794 reading features
2017-07-09 14:36:12,881 Merging features onto tracks
[...]

PR #217, opened 2017-09-13
Handle failed matching if at least one image has no GPS data

bdholt1:
Fix for #198 

The fix worked for me, don't see why everyone else using this library should need to apply the patch too.

### DIFF
--- a/opensfm/commands/match_features.py
+++ b/opensfm/commands/match_features.py
@@ -90,9 +90,14 @@ def match_candidates_by_distance(images, exifs, reference, max_neighbors, max_di
     points = np.zeros((len(images), 3))
     for i, image in enumerate(images):
         gps = exifs[image]['gps']
-        points[i] = geo.topocentric_from_lla(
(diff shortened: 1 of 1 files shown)


paulinus:
What should we do if one image does not has GPS and match selection using GPS distance is enabled?

The current behavior is to deactivate the GPS based match selection altogether.

If I understand correctly, with this patch we would keep it on with `max_distance=150`. The problem is that the distances between images with GPS and images without GPS will be arbitrarily long. So we will never select matching pairs between the two groups.

Does this make sense? What behavior would you want, @bdholt1 ?

markuswagner1:
To weigh in on this discussion as I have also had the same problem:

I consider the correct algorithmic behavior would be to include images that have invalid or missing GPS information in all matching pairs, hence, to match such an image with ALL the other images. If you have, as I do, usually a large amount of images with valid GPS information and just a few where this is missing, it would be a huge speedup compared to having all pairs evaluated.

Just had a quick glance into the code: I think it would be a rather minor patch to implement, I hope I'll have time next week to try it.

PR #223, opened 2017-10-11
Move features directory to trash on clean

lamerman:
Clean should return the data directory to the initial state.
Currently the features directory is is not cleaned. This patch
fixes that behaviour.


paulinus:
You are right, thanks. Will also remove the lines on that script using the old feature folder names.

Issue #190, opened 2017-06-30
Reconstruction process hangs indefinitely


pierotofy:
This is related to https://github.com/OpenDroneMap/WebODM/issues/147 and several other reports that we have encountered with OpenDroneMap users.

When processing large datasets, on computers with limited amount of RAM, while using a value of `processes` > 1 (forcing the use of `Pool` such as in https://github.com/mapillary/OpenSfM/blob/d22b94c6a46f5be88857290363943f084be4a7f9/opensfm/dense.py#L54), the process hangs indefinitely. This can be reproduced sporadically (not reliably) with real world inputs, but can be clearly simulated with this example. In this example, the `Killer` class acts to simulate a out-of-memory error, therefore sending a KILL signal to one of the subprocesses. 

Notice that the `Ended` string is never printed and the process never terminates.

```python
from multiprocessing import Pool
import time
import threading
from subprocess import check_output

def get_pid(name):
    return list(map(int,check_output(["pidof",name]).split()))

class Killer(threading.Thread):
    def __init__(self):
        threading.Thread.__init__(self)

    def run (self):
        print("Wait 3 seconds...")
        time.sleep(3)
        pids = get_pid('python')
        print(pids)
        print("Found " + str(len(pids)) + " python processes")
        print("Killing middle process")
        
        check_output(["kill", "-9", str(pids[int(len(pids) / 2)])])

        print("Processes killed, now waiting for termination...")

def f(x):
    time.sleep(6)
    return x*x

if __name__ == '__main__':
    p = Pool(3)
    print("Started")
    Killer().start()
    print(p.map(f, [1, 2, 3]))
    print("Ended")
```

I haven't investigated possible solutions, but ideally opensfm should report to the user that the process ran out of memory and exit gracefully.

paulinus:
Thanks for the minimal example!

From what I understand, the fact that signals are not handled when using `Pool` is because of a [python bug](https://bugs.python.org/issue8296). Similar problems and workarounds are reported in [here](https://stackoverflow.com/questions/1408356/keyboard-interrupts-with-pythons-multiprocessing-pool) and [here](https://stackoverflow.com/questions/11312525/catch-ctrlc-sigint-and-exit-multiprocesses-gracefully-in-python)

We should implement one of the workarounds since, currently, the only way of stopping the processes is using a sigquit or ^\ 

Apart from that, we were not terminating the Pools after using them in dense matching which could lead to a number of old processes hanging around. Should not happen anymore with current master.

[...]

PR #224, opened 2017-10-19
Replace multiprocessing with loky

pierotofy:
This PR replaces multiprocessing.Pool with loky.get_reusable_executor. The advantage of this is that processes that could have been killed by an out of memory problem will now cause the program to fail instead of letting it hang indefinitely. This has been affecting a lot of users of OpenDroneMap that are trying to process large datasets. #190 

List of changes:

- Added loky to the directory tree. Pulling in the library with `pip install loky` would remove the need for this addition, but there's no way to disable a warning on OSX that notifies the user that "semaphore are broken on OSX". This is a warning that can be safely dismissed (see https://github.com/tomMoral/loky/issues/103) in our case. I made a fork of the library and uncommented the warning (https://github.com/tomMoral/loky/compare/master...pierotofy:master), but didn't want to add my personal repository in the list of pip requirements. Perhaps the better way to go would be for mapillary to make a fork and then install this dependency via github. Using warnings.catch_warnings is not possible to intercept and dismiss the warning (see https://bugs.python.org/issue6647). If I need to change this, please let me know. Also, if we don't care about the warning messages, we could just install from directly from pip, OSX users will see the warning until we upgrade to Python 3.
- Replaced Pool with get_reusable_executor.
- Moved logging.basicConfig into a separate module (log.setup()); the reason for this is that get_reusable_executor needs logging.basicConfig to be set within the forked process, otherwise all logging activity disappears from the console.
- Added calls to log.setup() at the top of each function that's designed to be parallel and outputs some logs.


paulinus:
Thanks for doing this, @pierotofy.

I think it is better not to include loky code in OpenSfM. It would be a mess to maintain. Installing from your fork in github would be better but I think we can skip that too and just live with the warning.

Let's use default loky.

pierotofy:
Sounds good, thank you for reviewing these changes so quickly @paulinus !

Issue #165, opened 2017-04-16
Distortion coeficient up to K3 and additionally P1 and P2


specyfick:
I'm playing with OpenDroneMap. I'm using a highly distorted images from GoPro Hero 3 Black.
Unfortunately distortions coefficients are passed to algorithm for K1 and K2.
Undistortion results are rather poor ,due to the fact that to cv2.undistort only K1 and K2 are passed, P1 and P2 are assumed zero, K3 zero to.
I've calibrated images with other tool. Poor results with K1 and K2.
After I calibrated it with additionally P1 and P2 and K2 coefs, results are nice.
I will try to code something with opensfm in undistort.py, but I'm not good in it.
Please do it that way, that people can pass distortion coefs K1 K2 K3 and P1 and P2 from file to all sfm algorithm, when using their own cameras, not well defined in files.

Some results.
Distorted image:
![cal2_1](https://cloud.githubusercontent.com/assets/25548431/25071514/55b54170-22ba-11e7-88a3-0a066199b8c0.JPG)

Rectified image with K1 and K2, P1 and P2 coefs. Can see still radial in the center.
![cal2__rect1-k2](https://cloud.githubusercontent.com/assets/25548431/25071516/55b8ca20-22ba-11e7-8553-c082e9ac7baa.JPG)

Rectified image with K1 and K2 and **K3** P1 and P2 coefs additionally gamma (alpha skew) estimates. Can see radial distortion eliminated in the center.
![cal2__rect1](https://cloud.githubusercontent.com/assets/25548431/25071515/55b7fec4-22ba-11e7-9fbe-e69a99f18dc1.JPG)

As we estimate more coefs a model is better. There are more coef K4 and K5. The callibration would be even better.
Please consider this improvement.
Best regards.
Pawel

lamerman:
I vote for this issue, it's something to be done. For Galaxy S7 it's also impossible to undistort the image correctly without it.

lamerman:
I implemented, but has not yet uploaded the K3 addition and after this improvement "Final cost" of bundle adjustment decreased for one order of magnitude for Galaxy S7. The latest requires three parameters for calibration. 

@paulinus is this patch needed in general? I then would prepare it properly. Why I ask maybe someone is already working on it and there is no sense in mine.

paulinus:
[...]

PR #225, opened 2017-10-23
Add third parameter for radial distortion for perspective camera

lamerman:
Currently only K1 and K2 are used by the algorithm, while K3
is needed for certain cameras.

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -68,6 +68,7 @@ reprojection_error_sd: 0.004    # The startard deviation of the reprojection err
 exif_focal_sd: 0.01             # The standard deviation of the exif focal length in log-scale
 radial_distorsion_k1_sd: 0.01   # The standard deviation of the first radial distortion parameter (mean assumed to be 0)
 radial_distorsion_k2_sd: 0.01   # The standard deviation of the second radial distortion parameter (mean assumed to be 0)
+radial_distorsion_k3_sd: 0.01   # The standard deviation of the third radial distortion parameter (mean assumed to be 0)
 bundle_interval: 0              # bundle adjustment after adding 'bundle_interval' cameras
 bundle_new_points_ratio: 1.2    # bundle when (new points) / (bundled points) > bundle_outlier_threshold
 bundle_outlier_threshold: 0.006
--- a/opensfm/commands/undistort.py
+++ b/opensfm/commands/undistort.py
@@ -112,6 +112,7 @@ def perspective_camera_from_fisheye(fisheye):
     """Create a perspective camera from a fisheye."""
     camera = types.PerspectiveCamera()
     camera.id = fisheye.id
+    camera.distortion_model = 'radial_k2'
     camera.width = fisheye.width
     camera.height = fisheye.height
     camera.focal = fisheye.focal
@@ -123,6 +124,7 @@ def perspective_camera_from_fisheye(fisheye):
 def perspective_views_of_a_panorama(spherical_shot, width):
     """Create 6 perspective views of a panorama."""
     camera = types.PerspectiveCamera()
+    camera.distortion_model = 'radial_k2'
     camera.id = 'perspective_panorama_camera'
     camera.width = width
     camera.height = width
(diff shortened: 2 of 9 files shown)


lamerman:
So the idea here is that we introduce a concept of distortion model which currently can be:
- radial k1+k2
- radial k1+k2+k3
In the future I assume it should at least include tangential distortions (not the scope of this change).

Every perspective camera has a configurable distortion model. Fisheye as I understand also should have one, it's not yet implemented.

lamerman:
The difference in bundle adjustment score for galaxy s7 camera calibrated on checkerboard with 2 and 3 parameters is:
Final cost:  4.468651e+02 for 2 parameters
Final cost:  3.657365e+02 for 3 parameters

The scene contains 22 outdoor images.

I previously said the difference was order of magnitude in the issue #165, so it was not true, that measurements were not correct :)  The difference is less, but still present and seem to be stable.

paulinus:
That is great, @lamerman !

While we can indeed add the concept of "distortion model" we can also consider each distortion model as a new `projection_type`. We currently have `perspective`, `fisheye` and `equirectangular` projection models. We can add a 4th one that could be named `brown` and that would have the 5 distortion parameters of the Brown-Conrady distortion model.

That has the advantage that we can keep the current `perspective` model unchanged and we can freely play with the new `brown` model without breaking anything. The disadvantage is that we need to duplicate some code but i think it would be cleaner like this. What do you think?

lamerman:
@paulinus I think you are right, maybe it would be better to implement the full model (radial+tangential distortions) at once in a separate class. 

Right now I have some problems with time, but if I have some time window (hopefully) in the nearby future I will implement it. Please keep track in the task if someone has started to make it.

I'll close the PR now, as this code can't be reused for that implementation.

bdholt1:
Hi @lamerman and @paulinus - this is something that I'd like as well.  I'm happy to start on the implementation and help out. Perhaps we can work together on it?

lamerman:
@bdholt1 I'm not working on it now :) feel free to start.

bdholt1:
> While we can indeed add the concept of "distortion model" we can also consider each distortion model as a new projection_type. We currently have perspective, fisheye and equirectangular projection models. We can add a 4th one that could be named brown and that would have the 5 distortion parameters of the Brown-Conrady distortion model.

>That has the advantage that we can keep the current perspective model unchanged and we can freely play with the new brown model without breaking anything. The disadvantage is that we need to duplicate some code but i think it would be cleaner like this. What do you think?

If I understand @paulinus correctly, the proposal is to create a new class (in types.py)

`class BrownConradyPerspectiveCamera(Camera):`

which will have k3, p1 and p2 parameters; and a corresponding new projection_type 'brown'.

This will certainly result in some duplicated code (basically copy/paste PerspectiveCamera and add the new distortion coefficents) and it is 'cleaner' in the sense that if won't break existing code. The main drawback here is that any benefit gained by adding the new distortion coefficients will not be enjoyed by the Fisheye camera model (we'd need to create a new BrownConradyFisheyeCamera and projection type).

This PR started down the road of attempting to avoid that by introducing a 'distortion model' in the PerspectiveCamera but it's implementation only considered the PerspectiveCamera, so once again the benefits could not be enjoyed by the Fisheye camera.

An alternative would be to try to decouple the Camera and the Distortion model such that Camera 'has a ' distortion.  Intuitively this makes sense - they are independent concepts.

To decouple the concepts would look a bit like this current PR, but would introduce a new class in types.py called Distortion.We could then support different distortion models, for example (K1,K2) as is currently done as well as any number of new distortion models (K1,K2,P1,P2,K3) etc and then both Perspective and Fisheye cameras can have a Distortion.  The biggest changes will be in the bundle adjustment code.

Is it worth pursuing this idea?  The downside is that it's likely to impact existing code quite signficantly.

paulinus:
The ideal case would be that the current PerspectiveCamera already included those parameters and a way to choose which ones to optimize. We may end up there but I think it is easier to do it by adding a new camera model. If that camera model eventually gets flexible enough, it might replace the current perspective one. But I think it is a good idea to keep the current one untouched not to break existing reconstructions.

On separating the distortion model from the camera model, I don't think it is worth the effort. What is left on a camera model if you remove the distortion model?  A distortion model **is** a camera model.

Note for example that you can not build a distortion model that would work for both perspective and fisheye camera models.  While the models share the parameter names and a polynomial expression, the polynomial is applied to different radius concepts.

`<philosophy>`
We call it distortion because we think of it as a deviation from the ideal pinhole model. But there is another way to look at it that works for more camera models.

The goal of a (central) camera model is to map 3D points to 2D pixels and 2D pixels to 3D bearings.  These mappings may include a simple perspective projection or more involved, non-linear functions. As long as you can define the mapping, you have a camera. There is no need to define the mapping as a distortion from the pinhole model.
`</philosophy>`

So all in all, I would go for starting with a copy-paste of the current PerspectiveCamera and add the parameters there.

PR #229, opened 2017-10-27
add optional --no-camera when saving reconstruction ply

dakotabenjamin:
This will enable producing a sparse point cloud without the camera positions.


paulinus:
That is useful, thanks!
Will add a `no-points` option too.

PR #236, opened 2017-11-08
Adds Dense SIFT feature type. http://www.vlfeat.org/overview/dsift.html

sashichandra:


opensfm/src/dsift.cc
```diff
@@ -0,0 +1,52 @@
+#include "types.h"
+#include "dsift.h"
+
+#include <vector>
+#include <iostream>
+
+extern "C" {
+  #include "vl/dsift.h"
```
russkel: Are these files included in the pull req?

sashichandra:
Hi,
vl/dsift.h is already present in opensfm/src/third_party/vlfeat/vl/dsift.h
They are not part of this pull req.

-sashi.

russkel:
Hi @sashichandra, sorry I deleted my comment after I found the file.

russkel:
Is there a chance you could resolve the conflict in the features.py file @sashichandra ?

@paulinus are you willing to merge this? Otherwise I'll pull this into my own working copy.

paulinus:
@russkel, I've resolved the conflicts on this branch and put the result in this branch https://github.com/paulinus/OpenSfM/tree/sashichandra-dsift

About merging it to master, the main concern is that Dense SIFT is not a suitable feature for SfM in general.  The reason for that is that it extracts features on a fixed grid that is not adapted to the actual image texture. As a result, corresponding features in different images are not centered around the same real-world point.  This adds substantial error to triangulation and resection algorithms to the point where they no longer work.

Testing this branch on berlin and lund dataset, I was unable to get good reconstructions. Only berlin yield a reasonable reconstruction after setting `dsift_step` to a small number so that the discretization error was reduced. That produced a lot more features which were then slow to match.

If there is a trick to make things work better, please let me know.

Issue #202, opened 2017-07-21
Tips for achieving denser reconstructions


sola92:
Right now, the dense reconstructions are still pretty sparse; it seems like like not enough features are being extracted. What factors play into improving dense reconstructions. Any tips for how to improve this?

Data set for reference:
[dataset.zip](https://github.com/mapillary/OpenSfM/files/1164384/dataset.zip)

erikreed:
Have you tried exporting to openMVS and running `DensifyPointCloud` to produce a dense .ply? 

Pau has a nice writeup on it: http://blog.mapillary.com/update/2016/10/31/denser-3d-point-clouds-in-opensfm.html

paulinus:
For this example, the object has little texture. The way the pictures are taken is good, with lots of overlap, but since the object is shiny and textureless, there are few points that can be reconstructed.

PR #237, opened 2017-11-08
feature: add extra steps to yield a point cloud file

bdholt1:
Following #202 and http://blog.mapillary.com/update/2016/10/31/denser-3d-point-clouds-in-opensfm.html I found it valuable to always generate a point cloud at the end of the process.

### DIFF
--- a/bin/opensfm_run_all
+++ b/bin/opensfm_run_all
@@ -12,4 +12,3 @@ $DIR/opensfm reconstruct $1
 $DIR/opensfm mesh $1
 $DIR/opensfm undistort $1
 $DIR/opensfm compute_depthmaps $1
-$DIR/opensfm export_openmvs $1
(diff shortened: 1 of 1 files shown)


bin/opensfm_run_all
```diff
@@ -10,3 +10,6 @@ $DIR/opensfm match_features $1
 $DIR/opensfm create_tracks $1
 $DIR/opensfm reconstruct $1
 $DIR/opensfm mesh $1
+$DIR/opensfm undistort $1
+$DIR/opensfm compute_depthmaps $1
+$DIR/opensfm export_openmvs $1
```
paulinus: do we always want to export to `openmvs`?

bdholt1: Good point - no we don't.  But it is helpful at least to always generate the point cloud.

paulinus: Yes, it makes sense to have all the steps to get the point cloud here so they are visible. Thanks for adding this. If you can remove the last line, i'll merge.

PR #238, opened 2017-11-13
bug: account for fisheye camera type in reconstruction

bdholt1:
The fisheye camera parameters are not accounted for in the bundle adjustment.


paulinus:
good catch, thanks!

PR #239, opened 2017-11-13
feature: add camera type 'brown' to account for Brown-Conrady optical…

bdholt1:
Following the discussion in PR #225 here is a possible implementation of the 5 distortion parameters (Brown-Conrady) model that OpenCV also supports: K1, K2, P1, P2, K3.

This addresses Issue #165.

@lamerman @paulinus There is a problem somewhere in the bundle adjustment that causes this not to converge to a solution.  I've spent some time debugging it bug I can't seem to find the issue so I thought to put it up and ask because you both seem to have some experience with Ceres.

### DIFF
--- a/opensfm/src/bundle.h
+++ b/opensfm/src/bundle.h
@@ -18,6 +18,7 @@ extern "C" {
 
 enum BACameraType {
   BA_PERSPECTIVE_CAMERA,
+  BA_BROWN_PERSPECTIVE_CAMERA,
   BA_FISHEYE_CAMERA,
   BA_EQUIRECTANGULAR_CAMERA
 };
@@ -33,11 +34,15 @@ enum {
   BA_CAMERA_FOCAL,
   BA_CAMERA_K1,
   BA_CAMERA_K2,
-  BA_CAMERA_NUM_PARAMS
+  BA_CAMERA_BASIC_RADIAL_DISTORTION_NUM_PARAMS,
+  BA_CAMERA_P1 = BA_CAMERA_BASIC_RADIAL_DISTORTION_NUM_PARAMS,
+  BA_CAMERA_P2,
+  BA_CAMERA_K3,
+  BA_CAMERA_BROWN_DISTORTION_NUM_PARAMS
 };
 
 struct BAPerspectiveCamera : public BACamera{
-  double parameters[BA_CAMERA_NUM_PARAMS];
+  double parameters[BA_CAMERA_BASIC_RADIAL_DISTORTION_NUM_PARAMS];
   double focal_prior;
   double k1_prior;
   double k2_prior;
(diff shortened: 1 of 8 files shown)


bdholt1:
I have a small dataset with distortion coefficients which could be helpful to verify this PR. What's the best way to reference it? Shall I add it to data/ and update the PR ?

paulinus:
@bdholt1 it would be great if you can upload the data to a server and add the link here. Thanks!

About the convergence, it is likely that adding more parameters renders the problem ambiguous and hard to optimize. I think the main use case for using all of the brown parameters is to precalibrate the camera and then fix the parameters during bundle adjustment.

We can do that by setting a very small standard deviation on the parameters in `config.yaml`. We might want to add the possibility of not optimizing the parameters at all.

opensfm/src/bundle.h
```diff
@@ -51,8 +56,33 @@ struct BAPerspectiveCamera : public BACamera{
...
+struct BABrownPerspectiveCamera : public BACamera{
+  double parameters[BA_CAMERA_BROWN_DISTORTION_NUM_PARAMS];
+  double focal_prior;
+  double k1_prior;
+  double k2_prior;
+  double p1_prior;
+  double p2_prior;
+  double k3_prior;
+
+  BACameraType type() { return BA_PERSPECTIVE_CAMERA; }
```
paulinus: That should be BA_BROWN_PERSPECTIVE_CAMERA. I think that must be the reason why bundle was failing and you were not getting good results @bdholt1

paulinus:
@bdholt1, I finally took the time to look at this. I think everything was good except for the copy/paste error above. It works well in my limited experiments.

Since the master has diverged a little since this PR I've merged it manually, and it lives in this branch now https://github.com/mapillary/OpenSfM/compare/feature-brown-camera-model

I'll do some more tests and merge to master. Thanks for your effort!

bdholt1:
Thanks for reviewing @paulinus.  I feel a bit unlucky missing that one but
I'm glad you found it.  Do you need me to do anything else?  What's the
plan with the perspective camera in the future? Will it be removed seeing
as it's going to be redundant?

paulinus:
The simpler perspective model will not be redundant unless there is a simple way of locking a subset of the camera parameters. During bundle adjustment, if we let the parameters free, the result will not be the same with the two models since one has more free parameters than the others.

In general, optimizing all parameters of the Brown model during bundle adjustment is not a good idea since are generally underconstrained.

Thus, I see the following two scenarios

1. The exact calibration is unknown. In this case, the best is to use the simple perspective model and optimize its 3 parameters during bundle adjustment.

2. The exact calibration of the camera is known (for example using a chessboard calibration). In this case, the best is to use the Brown camera model and fix the parameters during bundle (using `optimize_camera_parameters: no`)

By the way, I've added the principal point parameters to the Brown model to make it equal to the one in OpenCV.

bdholt1:
OK, that makes sense to use the Brown model only when parameters are provided and use the perspective otherwise.  Is there anything else you need me to do?  I can have another go with the dataset to see how well it performs now that you've found the bug.

paulinus:
It would be great if you could try your dataset with the branch https://github.com/mapillary/OpenSfM/tree/feature-brown-camera-model
thanks!

lamerman:
Probably documentation should also be updated, for example `using.rst`,  the `extract_metadata`  part. Maybe some other docs too.

PR #244, opened 2017-11-29
Improve navigation for equirectangular images

### LINKED ISSUE #235: How to use OpenSfM viewer to watch 360 degree panorama photos
I would like to know how to use OpenSfM viewer to watch 360 degree panorama photos. like Google street map, I can look around, not only go forward and backward. I made the structure with equirectangular type photo. Many thx.

mpetroff:
This pull request improves navigation for equirectangular images by determining valid movement directions using the current viewing direction, which is also kept constant when moving between images. This should fix #235.

It would probably be better if the first image viewed had the camera horizontal and kept the same heading as was used in orbit mode, but I didn't see a straightforward way to detect that the moving mode was just changed.

### DIFF
--- a/viewer/reconstruction.html
+++ b/viewer/reconstruction.html
@@ -1074,7 +1074,15 @@
                 var reconstruction = camera.reconstruction;
                 var shot = reconstruction['shots'][camera.shot_id];
                 var cam = reconstruction['cameras'][shot['camera']];
-                controls.goto_shot(cam, shot);
+                if (cam.projection_type == 'equirectangular') {
+                    // Move camera to new location
+                    controls.animationPosition.copy(opticalCenter(shot));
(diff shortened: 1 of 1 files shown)


paulinus:
That is great @mpetroff !  Navigation for 360 was tedious.

I'm trying the code and have some problem with the arrow estimation. My guess is that by setting `turn=0` the computation of `driftBA` gets wrong.

I think that an alternative way of removing the turn check is to set `dir = currentDir` since the destination direction will be the same as the current one.
I've tried with
`var dir = cam.projection_type == 'equirectangular' ? currentDir : viewingDirection(shot);`
and then letting the original turn computation and it seems to do the trick.

paulinus:
I'll go ahead and merge this and apply the change after that. Thanks @mpetroff !

PR #252, opened 2017-12-14
Add feature detection and matching explaination

bryanibit:
Just add feature detection and matching explaination to doc/source, but not know how to build the markdown to html for now.


paulinus:
Thanks for doing this @bryanibit. We are going to do a small doc sprint soon and will get this merged.

Issue #254, opened 2017-12-19
Instead/along with printing transformation, be able to access it


dakotabenjamin:
https://github.com/mapillary/OpenSfM/blob/master/opensfm/commands/export_geocoords.py#L51-L54

I'd like to be able to get this either as a transformation matrix or in file form. Preferably the former. Can I add a return statement to run() or make transformation a global variable? 

EDIT:
I'm looking more into how the commands work and the way I'm trying to use this command may not be the best route for me. I will try to rework how I'm doing things and get back to this.

PR #255, opened 2017-12-19
Write geocoords transformation to file via #254

dakotabenjamin:
#254 This will create a transformation.txt file in the data base directory. It can then be parsed by other means.


paulinus:
Looks good, thanks! I'll add a parameter in case someones the output on a different file.

paulinus:
@dakotabenjamin, I've added an `--output` parameter and changed the default name to `geocoords_transformation.txt` to be more explicit about what it is. https://github.com/mapillary/OpenSfM/commit/bcffb387b1d41a0c82919c92fd38b07d1c2bed90

PR #258, opened 2017-12-22
python3 and boost 1.65 changes

spease:
This still doesn't work, but is some amount along. I'm just about done for the night and probably won't have any more time to work on this.

Note that this also tries to fix the Boost python numeric deprecation errors as well, with a slightly different (maybe more generic?) approach than I saw in opengv.

This isn't my usual cup of tea, so be extra skeptical.
```
2017-12-22 02:22:55,247 INFO: Extracting focal lengths for image 01.jpg
2017-12-22 02:22:55,290 INFO: Extracting focal lengths for image 03.jpg
2017-12-22 02:22:55,326 INFO: Extracting focal lengths for image 02.jpg
2017-12-22 02:22:56,513 INFO: Extracting ROOT_HAHOG features for image 02.jpg
2017-12-22 02:22:56,514 INFO: Extracting ROOT_HAHOG features for image 01.jpg
2017-12-22 02:22:56,559 INFO: Extracting ROOT_HAHOG features for image 03.jpg
2017-12-22 02:22:56,678 INFO: Found mask to apply for image 01.jpg
2017-12-22 02:22:56,680 INFO: Found mask to apply for image 02.jpg
2017-12-22 02:22:56,721 INFO: Found mask to apply for image 03.jpg
  File "/home/spease/projects/carlos/OpenSfM/bin/opensfm", line 34, in <module>
    command.run(args)
  File "/home/spease/projects/carlos/OpenSfM/opensfm/commands/detect_features.py", line 30, in run
    parallel_map(detect, arguments, processes)
  File "/home/spease/projects/carlos/OpenSfM/opensfm/context.py", line 38, in parallel_map
    return list(e.map(func, args))
  File "/home/spease/projects/carlos/OpenSfM/env/lib/python3.6/site-packages/loky/process_executor.py", line 769, in _chain_from_iterable_of_lists
    for element in iterable:
  File "/home/spease/projects/carlos/OpenSfM/env/lib/python3.6/site-packages/loky/_base.py", line 584, in result_iterator
    yield future.result()
  File "/home/spease/projects/carlos/OpenSfM/env/lib/python3.6/site-packages/loky/_base.py", line 431, in result
    return self.__get_result()
  File "/home/spease/projects/carlos/OpenSfM/env/lib/python3.6/site-packages/loky/_base.py", line 382, in __get_result
    raise self._exception
loky.process_executor.BrokenProcessPool: A process in the process pool was terminated abruptly while the future was running or pending.
```

### DIFF
--- a/opensfm/align.py
+++ b/opensfm/align.py
@@ -26,12 +26,12 @@ def apply_similarity(reconstruction, s, A, b):
     :param b: The translation vector (3)
     """
     # Align points.
-    for point in reconstruction.points.values():
+    for point in list(reconstruction.points.values()):
         Xp = s * A.dot(point.coordinates) + b
         point.coordinates = Xp.tolist()
 
     # Align cameras.
-    for shot in reconstruction.shots.values():
+    for shot in list(reconstruction.shots.values()):
         R = shot.pose.get_rotation_matrix()
         t = np.array(shot.pose.translation)
         Rp = R.dot(A.T)
@@ -67,7 +67,7 @@ def align_reconstruction_naive_similarity(reconstruction, gcp):
         Xp.extend(measured)
 
     # Get camera center correspondences
-    for shot in reconstruction.shots.values():
+    for shot in list(reconstruction.shots.values()):
         X.append(shot.pose.get_origin())
         Xp.append(shot.metadata.gps_position)
 
@@ -103,7 +103,7 @@ def align_reconstruction_orientation_prior_similarity(reconstruction, config):
     X, Xp = [], []
     orientation_type = config['align_orientation_prior']
     onplane, verticals = [], []
-    for shot in reconstruction.shots.values():
+    for shot in list(reconstruction.shots.values()):
         X.append(shot.pose.get_origin())
         Xp.append(shot.metadata.gps_position)
         R = shot.pose.get_rotation_matrix()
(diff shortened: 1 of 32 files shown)


paulinus:
That's great, @spease. Thanks for getting this started. It will take some time for us to be able to merge this to master but we are really looking forward to starting python3 support.

Quick question: are the changes in this PR intended to make it work with python3 only or on both 2 and 3?

paulinus:
For boost 1.65 and opengv, I've started playing with it here https://github.com/laurentkneip/opengv/compare/master...paulinus:boost-python-numpy
Will compare that to the approach you took here.

spease:
Only python3 - I primarily ran the 2to3 tool on python scripts and then manually fixed some errors relating to the Json encoding. Nb that by fixed I mainly mean removing it, so far everything had used utf-8 encoding and it didn't look like there was an easy way to specify since the Json library writes strings and the encoding argument looked like it was being passed after files were being opened (best option would probably be to set the file encoding).

Build.py is probably compatible with both from me including the future import.

Note that the cmakelists probably needs more work to work with different platforms and versions, since some platforms use boost-python_py36 instead of just boost python3 from reading about other conversation efforts out there. I saw tickets both for a fallback in that case, and a cmake ticket to make things pick the correct variant based on just a single specifier.

paulinus:
We have added initial python 3 support maintaining python 2 compatibility. Thanks for paving the way @spease !

PR #280, opened 2018-01-30
vectorized version of back_project

mlopezantequera:
I'm not sure if the naming conventions are the best.


paulinus:
It is clearly better than the trailing `s` in `pixel_bearings`

PR #286, opened 2018-02-08
Angular threshold

mlopezantequera:
This needs reviewing as it changes the default behavior of robust_match_calibrated:

Before:
- ransac was run with a threshold of 0.004 (default value of
  robust_matching_threshold)
- inliers were filtered with a threshold of 0.01 (fixed)

After:
- ransac and inliers use a threshold of 0.01 (default value of a new
  parameter: robust_matching_calib_threshold)

It depends on #285 and branch 'vectorizations'


paulinus:
Much cleaner now like this. Thanks @mlopezantequera!

PR #296, opened 2018-03-09
allow openSfm accept  proposed ODM  gcp_list.txt with comments. ODM  …

yjmenezes:
Allow openSfm accept  proposed ODM  gcp_list.txt with comments. ODM   PR #764
Comments will turn gcp_list.txt more flexible. Turn GCP on/off.


paulinus:
That will be more flexible, thanks!

PR #307, opened 2018-04-11
Fix exception caused by no EXIF focal length

russkel:
This fixes the exception caused when there is no exif/focal length data for the image.


paulinus:
perfect, thanks!

PR #309, opened 2018-04-15
Fixed GCP UTM south region parser typo

pierotofy:
A typo is causing GCP files with a UTM projection in the southern regions (ex. `WGS84 UTM 32S`) to fail with `zone_number = int(zone['-1']) TypeError: string indices must be integers, not str`.

First reported here: http://community.opendronemap.org/t/problem-using-a-gcp-file-in-the-southern-hemisphere-wgs84-utm-36s/495


paulinus:
ups. Thanks for catching and fixing!

PR #316, opened 2018-04-21
Fix+small clean up of plot scripts

russkel:
These scripts now run on python3, plus removal of unneeded argument in match_candidates_by_order function.


paulinus:
Thanks for fixing this @russkel.

Cool to see that you builded opensfm for python3. Please do share any issues or uncoviniences with the python3 installation.

PR #327, opened 2018-05-22
'\n'--> u'\n' for io.py write to fix string errors'

PeterSprague:


paulinus:
Thanks for finding this @PeterSprague. After the python3 compatibility changes there are a number of strings that need to be unicode spread though the code base. I've just imported `unicode_literals` now so we do not need to explicitly tag Unicode strings as such. Closing this PR since the explicity `u''` is not needed after that change.

PeterSprague:
This error is still appearing using the latest OpenSfM master when OpenDroneMap is built and run on a J3455 processor. Interestingly, a HP amd64 workstation does not produce the error.

PeterSprague:
Was my thought.  I will check that the J3455 SBC has the same OpenSfM version as workstation, though completely built OpenDroneMap from scratch on it this week.  That included OpenSfM master.

Anything else I should check?

PR #330, opened 2018-06-04
Specify UTF in writelines

smathermather:
Via @hhovde

### DIFF
--- a/opensfm/io.py
+++ b/opensfm/io.py
@@ -653,12 +653,12 @@ def export_bundler(image_list, reconstructions, track_graph, bundle_file_path,
         bundle_file = os.path.join(bundle_file_path,
                                    'bundle_r' + str(j).zfill(3) + '.out')
         with open_wt(bundle_file) as fout:
-             fout.writelines('\n'.join(lines) + '\n') => fout.writelines(unicode('\n'.join(lines) + '\n', 'utf-8'))
+             fout.writelines(unicode('\n'.join(lines) + '\n', 'utf-8'))
 
         list_file = os.path.join(list_file_path,
                                  'list_r' + str(j).zfill(3) + '.out')
         with open_wt(list_file) as fout:
-             fout.writelines('\n'.join(map(str, image_list))) => fout.writelines(unicode('\n'.join(map(str, image_list)),'utf-8'))
+             fout.writelines(unicode('\n'.join(map(str, image_list)),'utf-8'))
 
 
 def import_bundler(data_path, bundle_file, list_file, track_file,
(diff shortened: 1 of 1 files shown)


opensfm/io.py
```diff
@@ -653,12 +653,12 @@ def export_bundler(image_list, reconstructions, track_graph, bundle_file_path,
...
                                    'bundle_r' + str(j).zfill(3) + '.out')
         with open_wt(bundle_file) as fout:
-            fout.writelines('\n'.join(lines) + '\n')
+             fout.writelines('\n'.join(lines) + '\n') => fout.writelines(unicode('\n'.join(lines) + '\n', 'utf-8'))
 
         list_file = os.path.join(list_file_path,
                                  'list_r' + str(j).zfill(3) + '.out')
         with open_wt(list_file) as fout:
-            fout.writelines('\n'.join(map(str, image_list)))
+             fout.writelines('\n'.join(map(str, image_list))) => fout.writelines(unicode('\n'.join(map(str, image_list)),'utf-8'))
```
hhovde: fout.writelines(unicode('\n'.join(map(str, image_list)),'utf-8'))

paulinus:
@hhovde, do you know why we need to convert to unicode from utf here?
I've recently changed this file so that all string literals are unicode. After this change `'\n'.join(lines) + '\n'` should already be a unicode string. I'm afraid that converting it again to unicode can be decoding things twice.

hhovde:
@paulinus 
No I do not know, I haven't had the time to study the codebase.
This was a quickfix that resolved an issue while trying to get the latest OSfM to work with ODM. After pulling the latest OSfM and ODM, I got the following error:

`
File "OpenDroneMap/SuperBuild/src/opensfm/bin/export_bundler", line 48, in <module>
io.export_bundler(images, reconstructions, track_graph, bundle_file_path, list_file_path)
File "OpenDroneMap/SuperBuild/src/opensfm/opensfm/io.py", line 652, in export_bundler
fout.writelines('\n'.join(lines) + '\n')
TypeError: write() argument 1 must be unicode, not str
File "OpenDroneMap/run.py", line 47, in <module>
plasm.execute(niter=1)
File "OpenDroneMap/scripts/run_opensfm.py", line 173, in process
(context.pyopencv_path, context.opensfm_path, tree.opensfm))
File "OpenDroneMap/opendm/system.py", line 34, in run
raise Exception("Child returned {}".format(retcode))
Exception: Child returned 1
`
As I said I haven't had the time to read the code, so I don't know where or when the string data is introduced, but it might be from somewhere in the ODM code, in wich case that would be the better place to apply a patch.

smathermather:
I'm going to close this. We'll revisit if we need to when we pull in a later OpenSfM.

PR #333, opened 2018-06-05
Add class to reconstructed points when segmentation is available

paulinus:


silence401:
hi, i have some question. I have color image and segmentation image and I want to reconstruction points with label. Can opensfm do it  and how should I do.

paulinus:
opensfm will work with the color images to build a reconstruction and a dense point cloud. The dense point cloud will include a label for every point derived from the segmentation images.

silence401:
I got it ,thans for your reply

PR #348, opened 2018-08-07
Show different reconstructions with different colors

YanNoun:
This PRs makes the viewer display each reconstruction with a different pseudo-random color, for easier debugging purposes. We also display the reconstruction index below the thumbnail. Since Math.random() doesn't provide seed initialization, a pseudo-random-generator has been implemented, for reproducibility purposes.

### DIFF
--- a/viewer/reconstruction.html
+++ b/viewer/reconstruction.html
@@ -772,7 +772,7 @@
                 var s = 1
                 if(id !== 0){
                     s = Math.sin(id) * 10000; 
-                    s - Math.floor(s);
+                    s -= Math.floor(s);
                 }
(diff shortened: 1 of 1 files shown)


viewer/reconstruction.html
```diff
@@ -760,6 +768,16 @@
                 return rotation;
             }
 
+            function reconstruction_color(id){
+                var s = 1
+                if(id !== 0){
+                    s = Math.sin(id) * 10000; 
+                    s - Math.floor(s);
```
paulinus: Do you want a `-=` here?

YanNoun: Yes

PR #352, opened 2018-08-09
Added ability to undistort to PNG, scale images

pierotofy:
Hello :hand:

This PR proposes a modification to the undistort command to optionally allow a user to export images in PNG format and optionally scale the images before saving them to disk.

It also modifies the export_visualsfm command to optionally allow a user to choose the extension that the NVM file should reference (to be used in combination with the undistort command) and optionally scale the focal length value (which is useful if we have resized the images to perform further processing / visualization with external programs).

Other export commands haven't been modified, but if I need to modify them as well to provide better consistency, just ask, I'd be happy to.

```
usage: opensfm undistort [-h] [--image_format {jpg,png}]
                         [--image_scale IMAGE_SCALE]
                         dataset
usage: opensfm export_visualsfm [-h] [--undistorted]
                                [--image_extension {jpg,png}]
                                [--scale_focal SCALE_FOCAL]
                                dataset
```
The behavior of both programs is unchanged if the user leaves the defaults.

I hope these changes can be helpful to other users. If the proposed changes don't fit into the scope of OpenSfM I can always create a separate module, but I thought I'd share the work to see if there was interest in merging this in.

Thank you!

### DIFF
--- a/opensfm/commands/undistort.py
+++ b/opensfm/commands/undistort.py
@@ -20,6 +20,8 @@ class Command:
 
     def add_arguments(self, parser):
         parser.add_argument('dataset', help='dataset to process')
+        parser.add_argument('--image_format', choices=["jpg", "png"], default="jpg", type=str, help='Use this format to export images. Default: %(default)s')
+        parser.add_argument('--image_scale', default=1.0, type=float, help='Scale exported images by this factor. Default: %(default)s (no scale)')
 
     def run(self, args):
         data = dataset.DataSet(args.dataset)
@@ -27,11 +29,11 @@ class Command:
         graph = data.load_tracks_graph()
 
         if reconstructions:
-            self.undistort_reconstruction(graph, reconstructions[0], data)
+            self.undistort_reconstruction(graph, reconstructions[0], data, args.image_format, args.image_scale)
 
         data.save_undistorted_tracks_graph(graph)
 
-    def undistort_reconstruction(self, graph, reconstruction, data):
+    def undistort_reconstruction(self, graph, reconstruction, data, image_format = "jpg", image_scale = 1):
         urec = types.Reconstruction()
         urec.points = reconstruction.points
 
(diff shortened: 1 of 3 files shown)


paulinus:
Hi @pierotofy, thanks for the changes.
Can I ask what is the motivation for scaling the undistorted images? Is it for reducing the file size?

opensfm/commands/undistort.py
```diff
@@ -121,13 +130,25 @@ def undistort_image(arguments):
...
 
+def scale_image(image, scale_factor):
+    """Scale an image by a factor."""
+    if scale_factor == 1.0:
+        return image
+    
+    height, width, _ = image.shape
+    width = int(width * scale_factor)
+    height = int(height * scale_factor)
+    return cv2.resize(image, (width, height), interpolation=cv2.INTER_NEAREST)
```
paulinus: For downsampling, we should use `cv2.INTER_AREA` to reduce aliasing artifacts.

pierotofy:
> Hi @pierotofy, thanks for the changes.
> Can I ask what is the motivation for scaling the undistorted images? Is it for reducing the file size?

Yes. In the case of OpenDroneMap, when images are used for texturing our meshes, having smaller undistorted images can drastically reduce memory usage as well.

opensfm/commands/undistort.py
pierotofy: For the purpose of texturing meshes, in my tests I've tried both. Using cv2.INTER_AREA I noticed that the images are obviously smoother, but adds significant blur in the final textured mesh. I guess the best choice would depend on the use case. I could add an extra flag to specify what interpolation method to use?

paulinus:
Ok, thanks. I would consider using jpg instead of png when file size is a concern.

One thing to keep in mind is that undistorted images are used for dense matching to produce the point clouds. Changing the size or interpolation used to generate them can affect the point cloud results.

Let's merge and use this options carefully.

opensfm/commands/undistort.py
paulinus: No need for an option there. It would add complexity and we can pass without it.

pierotofy:
Thanks @paulinus!

paulinus:
@pierotofy, I had to revert the merge because it broke the undistortion for datasets with masks or segmentations.
I'll reactivate the test on the undistortion command so that this gets detected earlier.

paulinus:
@pierotofy, @smathermather, I finally got to look at this. I've updated the code so that test pass. Also moved the command line options to config options to avoid different commands using different options.

Could you please look at PR #380 and check if it would suit WebODM needs? thanks

smathermather:
Happy to test, but first to check with @pierotofy -- you could speak better to the API changes (i.e. command line options vs. config). Does this trigger changes in ODM?

PR #358, opened 2018-08-27
Points projections & Viewer features

YanNoun:
This PR adds the Measurement type : a measured quantity can be a 2D keypoint, a GPS position, IMU angle or surveyed GCP. In this PR, we did the following :

- Use Measurement type for describing 2D keypoints of reconstructed 3D points, that further allow for a complete scene geometry description.
 - Added point filtering by their # of projection using the above.
 - Added display of the camera connection graph using the first.

As a down-side, it makes the JSON files much heavier : Firefox might struggle at read > 200 Mb files (but Chrome can).

### DIFF
--- a/opensfm/io.py
+++ b/opensfm/io.py
@@ -126,9 +126,7 @@ def measurement_from_json(obj):
     """
     Read a measurement from a json object
     """
-    return types.Measurement(obj["coordinates"],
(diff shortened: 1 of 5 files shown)


YanNoun:
Closing, will not be merged.

PR #367, opened 2018-10-22
Enable overriding of default OpenMP number of threads limit in libAKAZE

GITNE:
`OMP_MAX_THREADS` can now be specified without having to modify `AKAZEConfig.h`. This also serves as a safeguard against accidental check-in of a modified `AKAZEConfig.h`, which prevents messing up the default configuration.


paulinus:
thanks, @GITNE

PR #369, opened 2018-10-24
feature: Extract true sensor width and focal length from EXIF

GITNE:
Most dedicated cameras report the focal length and sensor size in the EXIF header. Traditionally, only few smartphones have been reporting this data but with the latest smartphones on the market, more and more devices report this crucial data too. This data should finally enable correct camera calibration and ultimately also improve reconstruction quality (mainly precision with increasing distance).

For this feature to work the EXIF header must have the `FocalLength`, `FocalPlaneXResolution`, and `FocalPlaneResolutionUnit` tags. Meta data extraction still falls back to database values on known cameras which do not report the required EXIF tags or to the default focal ratio (`default_focal_prior`) if the camera could not be found in the database. Please note that many of the sensor size values in the database are skewed or copied from marketing material instead of technical documentation. Anyhow, the values in the EXIF header should always be preferred over any database values.

### DIFF
--- a/opensfm/exif.py
+++ b/opensfm/exif.py
@@ -11,6 +11,7 @@ from opensfm import types
 
 logger = logging.getLogger(__name__)
 
+inch_in_mm = 1609334.0 / 1760.0 / 3.0 / 12.0
 
 def eval_frac(value):
     try:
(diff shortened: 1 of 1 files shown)


GITNE:
Oh, I forgot to mention that finally *`unknown`* cameras, like new ones or which simply are not in the database, get the correct focal ratio as long as there is proper EXIF data in the image file.

GITNE:
@paulinus Ping? I am aware that you folks are quite busy but I think that this PR is also quite straight forward and does not break anything. If you seem to be too busy, my tip would be to try to find some time for PR review at least once a week, half a day or so. Well, this used to work for me :wink:

opensfm/exif.py
```diff
@@ -176,10 +181,25 @@ def extract_focal(self):
         focal_35, focal_ratio = compute_focal(
             get_tag_as_float(self.tags, 'EXIF FocalLengthIn35mmFilm'),
             get_tag_as_float(self.tags, 'EXIF FocalLength'),
-            get_tag_as_float(self.tags, 'EXIF CCD width'),
+            self.extract_sensor_width(),
```
paulinus: By doing this, we are no longer reading the `EXIF CCD width` tag.  Should `extract_sensor_width()` handle the `EXIF CCD width` tag?

paulinus: Awesome, wonder how we got look for it. Was just checking now and could not find any reference of it anywhere. Thanks for looking at this.

paulinus:
@GITNE, i've refactored the function to accept also other units https://github.com/mapillary/OpenSfM/commit/a074feab193776569ff23ce0ff5cab31535e34ff
I've followed the unit definitions in https://www.sno.phy.queensu.ca/~phil/exiftool/TagNames/EXIF.html
I do not have image examples using other units to test it on. Please, let me know if that looks correct to you. Thanks.

GITNE:
+1 Alright, looks good to me.

PR #371, opened 2018-10-26
Add time zone support to time stamp extraction

GITNE:
`DateTime*` time stamps are adjusted by time zone offsets `OffsetTimeOriginal`, `OffsetTimeDigitized`, and `OffsetTime` respectively, if available. Otherwise, UTC is naively assumed on any `DateTime*`.

The time stamp is extracted in the following order of decreasing precedence:
 * `GPSDateStamp` and `GPSTimeStamp`
 * `DateTimeOriginal` + `SubSecTimeOriginal` + `OffsetTimeOriginal`
 * `DateTimeDigitized` + `SubSecTimeDigitized` + `OffsetTimeDigitized`
 * `DateTime` + `SubSecTime` + `OffsetTime`
 * `DateTimeOriginal` + `SubSecTimeOriginal` (*UTC*)
 * `DateTimeDigitized` + `SubSecTimeDigitized` (*UTC*)
 * `DateTime` + `SubSecTime` (*UTC*)

### DIFF
--- a/bin/focal_from_exif
+++ b/bin/focal_from_exif
@@ -59,4 +59,4 @@ data.save_camera_models(camera_models)
 
 end = time.time()
 with open(data.profile_log(), 'a') as fout:
-    fout.write('focal_from_exif: {0}\n'.format(end - start))
+    fout.write('extract_metadata: {0}\n'.format(end - start))
--- a/opensfm/commands/extract_metadata.py
+++ b/opensfm/commands/extract_metadata.py
@@ -57,7 +57,7 @@ class Command:
 
         end = time.time()
         with open(data.profile_log(), 'a') as fout:
-            fout.write('focal_from_exif: {0}\n'.format(end - start))
+            fout.write('extract_metadata: {0}\n'.format(end - start))
 
     def _extract_exif(self, image, data):
          # EXIF data in Image
--- a/opensfm/exif.py
+++ b/opensfm/exif.py
@@ -28,11 +28,17 @@ def gps_to_decimal(values, reference):
     return sign * (degrees + minutes / 60 + seconds / 3600)
 
 
-def get_tag_as_float(tags, key):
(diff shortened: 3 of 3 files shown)


GITNE:
@paulinus This PR is probably not absolutely required for OpenSfM to work, since OpenSfM only depends on proper relative ordering of images, not absolute ordering. Anyhow, correct absolute time stamp extraction may be useful for some users, including Mapillary, although I am not sure at which stage of the Mapillary backend a time stamp is extracted. It would be nice to finally have this *feature* on the website. :wink:

GITNE:
> I am not sure at which stage of the Mapillary backend pipeline a time stamp is extracted.

@peterneubauer Perhaps you can comment on this. Currently, I am still very upset that time zone support does not work as expected…

peterneubauer:
Our precendence in the Mapillary backend currently is this:

```
"EXIF DateTimeOriginal",
"Image DateTimeOriginal",
"EXIF DateTimeDigitized",
"Image DateTimeDigitized"
"GPSTime"
"Image DateTime",
"EXIF DateTime"
```

@paulinus do you think it makes sense to change to the same order as suggested here?

GITNE:
> Our precendence in the Mapillary backend currently is this:
> ```
> "Image DateTimeOriginal",
> "Image DateTimeDigitized",
> "Image DateTime"
> ```
@peterneubauer I am not sure what you mean by *`Image`*?

The reason why I would go with `GPSTimeStamp` first is because it is the only time zone aware time stamp ever since EXIF 1.0. So, this should always work, even though not all devices write this tag. Then, we basically agree on the general order of precedence of `DateTime*` tags. However, time zone offsets denoted by `OffsetTime*` (if available) should be applied on `DateTime*` tags before finally falling back to and naively assuming UTC.
Changing the precedence on the backend and updating the database on all images *should* *not* be very expensive, since this is an extremely cheap operation, even though it may take some time.

PR #375, opened 2018-10-31
Update Installation Instructions

josiahls:
Changed: `git clone https://github.com/paulinus/opengv.git` to `git clone --recurse-submodules -j8 https://github.com/paulinus/opengv.git`

Notes:
opengv contains sub modules, we need recursive git cloning, otherwise not all the modules will be loaded.


paulinus:
Thanks!

PR #384, opened 2018-11-29
Tiny fix to allow blanks in image_name field in GCP file

markuswagner1:
Just a tiny fix as I was dealing with image filenames that had blanks


paulinus:
Thanks for point it out, @markuswagner1. We have fixed the issue by only splitting 5 times so that we do not split the file name in the first place (https://github.com/mapillary/OpenSfM/commit/b322c1004c2da8890b1b7090bb1d8a1de1809afc).  Please, let me know if that works for you. Thanks.

PR #394, opened 2019-02-01
Changed GPS based matching:

markuswagner1:
- If an image does not have GPS information in Exif, it is matched with ALL
  other images. A warning is written in the logs when that happens.

### DIFF
--- a/opensfm/commands/match_features.py
+++ b/opensfm/commands/match_features.py
@@ -105,19 +105,26 @@ def match_candidates_by_distance(images, exifs, reference, max_neighbors, max_di
 
     points = np.zeros((len(images), 3))
     for i, image in enumerate(images):
-        gps = exifs[image]['gps']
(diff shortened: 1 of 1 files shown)


paulinus:
Thank you for doing this, @markuswagner1. That is a better behavior, and some users already asked for this feature.

There is a corner case to solve in the current version:

The images that do not have GPS are letting the `points` array with 0 values. These points are still used later when we search for the neighbors of the images that do have GPS. This means that images with GPS values close to 0 can get neighbors that do not have GPS values instead of getting their real neighbors.

A solution would be to start by splitting the image list into two: one with images that have GPS and one with images that don't. Then we can process as before the images that have GPS and add all neighbors to the ones that don't.

markuswagner1:
Thanks @paulinus for reviewing and I agree that the corner case needs to be addressed. However, I am assigned to a different project, so I cannot carry it out in the next time.

PR #395, opened 2019-02-03
Resolve export_openmvs.py Python 3 compat issue.

mribbons:
It looks like other uses of .keys() are just wrapped in list(), let me know if you would prefer that for this line also.

Test with:
```sh
bin/opensfm_run_all data/berlin
bin/opensfm export_openmvs data/berlin
```

Error:
```
  File "bin/opensfm", line 34, in <module>
    command.run(args)
  File "/home/user/OpenSfM_python3/opensfm/commands/export_openmvs.py", line 26, in run
    self.export(reconstructions[0], graph, data)
  File "/home/user/OpenSfM_python3/opensfm/commands/export_openmvs.py", line 57, in export
    exporter.add_point(coordinates, shots)
TypeError: add_point(): incompatible function arguments. The following argument types are supported:
    1. (self: opensfm.csfm.OpenMVSExporter, arg0: numpy.ndarray[float64], arg1: list) -> None

```

### DIFF
--- a/opensfm/commands/export_openmvs.py
+++ b/opensfm/commands/export_openmvs.py
@@ -49,6 +49,11 @@ class Command:
 
         for point in reconstruction.points.values():
             shots = graph[point.id].keys()
+
+            # Python 3: keys() returns a dict_keys, OpenMVSExporter::add_point() requires a list
+            if not isinstance(shots, list):
+                shots = list(shots)
+
             coordinates = np.array(point.coordinates, dtype=np.float64)
             exporter.add_point(coordinates, shots)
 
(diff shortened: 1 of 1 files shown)


opensfm/commands/export_openmvs.py
```diff
@@ -49,6 +49,11 @@ def export(self, reconstruction, graph, data):
 
         for point in reconstruction.points.values():
             shots = graph[point.id].keys()
```
paulinus: Thanks for catching this, @mribbons.
What do you think about this version, which would work for both python 2 and 3
```suggestion
            shots = list(graph[point.id])
```

mribbons:
Hi @paulinus 
Tested your changes and the work for Python2 and 3. Got rid of the extra type check.

paulinus:
awesome, thanks!

PR #417, opened 2019-05-03
feature: scale 2D projection errors by their detected scale

YanNoun:
This PR makes 2D projection errors being scaled by their respective detected scale. Improves results on Strecha's bencmark by 25-50%.

A minor issue is that the feature were always exported by their actual pixel scale instead of a normalized one : hence a normalization hack.

Found a general issue in the bundle code, where errors should by scaled by `sqrt(1/std_deviation)` instead of just `1/std_deviation`

### DIFF
--- a/opensfm/dataset.py
+++ b/opensfm/dataset.py
@@ -733,7 +733,7 @@ class DataSet(object):
         return self.load_mask(image)
 
 
-TRACKS_VERSION = 0
+TRACKS_VERSION = 1
 TRACKS_HEADER = u'OPENSFM_TRACKS_VERSION'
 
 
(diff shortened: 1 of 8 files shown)


opensfm/dataset.py
```diff
@@ -735,12 +735,13 @@ def mask_as_array(self, image):
 def load_tracks_graph(fileobj):
     g = nx.Graph()
     for line in fileobj:
-        image, track, observation, x, y, R, G, B = line.split('\t')
+        image, track, observation, x, y, scale, R, G, B = line.split('\t')
```
YanNoun: Add some basic versioning

PR #419, opened 2019-05-22
feat:scale properly the feature scale

YanNoun:
This PR aims at correcting some hack introduced here (https://github.com/mapillary/OpenSfM/pull/417) : we want the feature scale to be properly set, at extraction time, when pixel position are normalized wrt. the image size.

To do so, we  :
 - Moved features IO to the `features` module
 - Added basic versioning. In the 0 version, normalized scale is set to the default config value (`0.004`)
 - Removed the normalizing hack in `tracking`
 - Added proper normalization in `features`

### DIFF
--- a/opensfm/features.py
+++ b/opensfm/features.py
@@ -313,3 +313,27 @@ def build_flann_index(features, config):
                         iterations=config['flann_iterations'])
 
     return context.flann_Index(features, flann_params)
+
+
+def load_features(filepath, config):
+    feature_type = config['feature_type']
+    s = np.load(filepath)
+    if feature_type == 'HAHOG' and config['hahog_normalize_to_uchar']:
+        descriptors = s['descriptors'].astype(np.float32)
+    else:
+        descriptors = s['descriptors']
+    return s['points'], descriptors, s['colors'].astype(float)
+
+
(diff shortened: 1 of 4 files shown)


opensfm/features.py
```diff
@@ -313,3 +315,64 @@ def build_flann_index(features, config):
...
+
+
+FEATURES_VERSION = 1
+FEATURES_HEADER = 'OPENSFM_FEATURES_VERSION'
+
+
+def load_features(filepath, config):
+    """ Load features from filename """
+    s = np.load(filepath)
+    version = _features_file_version(s)
```
paulinus: Should we get the version from s['OPENSFM_FEATURES_VERSION'] instead?

YanNoun: It's what `_features_file_version` does. I do agree that I use a var named FEATURES_HEADER that contains the string `OPENSFM_FEATURES_VERSION`, which is a bit conflated. Let me know if you prefer direct use of `OPENSFM_FEATURES_VERSION` instead.

paulinus: ok, i see. I got confused by the name HEADER and though the function was looking for a file header.
looks good

PR #421, opened 2019-05-30
FIX: cast to float in undistort.py

pierotofy:
Problem: when setting `undistorted_image_max_size` to a value lower than the images (eg. 100) the value of:

`factor = max_size / max(height, width)`

Always evaluates to 0 during undistortion.

Proposed solution: type cast to float the denominator.


paulinus:
This also had different behavior in python 2 vs python 3. Thanks for the fix.

PR #422, opened 2019-05-30
Faster export to VisualSFM

pierotofy:
Currently calling `export_visualsfm` will read the entire dataset of images from disk to compute width and height for each image.

This can be sped up a lot by simply reading the header values instead of loading the entire image in memory.

Even on a small dataset (a few dozen images):

```
time opensfm/bin/opensfm export_visualsfm /datasets/test

real	0m9.027s
user	0m8.164s
sys	0m1.302s
```

Vs. the proposed change:

```
time opensfm/bin/opensfm export_visualsfm /datasets/test

real	0m0.802s
user	0m0.694s
sys	0m0.500s
```

I'm not sure if the function to compute the width/height should move moved elsewhere? Anyway, let me know if anything needs to be changed :pray: Thank you!

### DIFF
--- a/opensfm/commands/export_visualsfm.py
+++ b/opensfm/commands/export_visualsfm.py
@@ -63,7 +63,8 @@ class Command:
         try:
             file_path = data._undistorted_image_file(image)
             with Image.open(file_path) as img:
-                return img.size
+                width, height = img.size
+                return height, width
         except:
             # Slower fallback
             image = data.load_undistorted_image(image)
--- a/opensfm/commands/export_visualsfm.py
+++ b/opensfm/commands/export_visualsfm.py
@@ -2,6 +2,7 @@ from __future__ import unicode_literals
 
 import logging
 import os
+import struct
 
 from opensfm import dataset
 from opensfm import transformations as tf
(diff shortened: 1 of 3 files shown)


paulinus:
That's quite a speed up!

It seems that Pillow can get the image size without decompressing the pixels https://stackoverflow.com/a/34704661/1380536

That would be a simpler solution that would not require adding the self-contained image size code.

pierotofy:
Oh, interesting. Did not know about Pillow. Will look into it and replace it if it works.

pierotofy:
That would require adding Pillow as a dependency however. https://github.com/mapillary/OpenSfM/blob/master/requirements.txt

If you're OK adding it, let me know.

pierotofy:
Same dataset benchmark with Pillow code:

```
real	0m11.229s
user	0m9.086s
sys	0m1.661s
```

So close to equivalent.

paulinus:
> Same dataset benchmark with Pillow code:
> 
> ```
> real	0m11.229s
> user	0m9.086s
> sys	0m1.661s
> ```

This looks slower than the other options. Are these Pillow times?
Otherwise, it is fine to add Pillow as a dependency if needed.

pierotofy:
Ups! Sorry it was late at night when I looked at the Pillow numbers. :laughing: they are actually much slower, like you say. So it seems that Pillow is reading the images in memory as well and the current code performs much faster.

paulinus:
Not sure what went wrong. Doing a quick test, Pillow seems orders of magnitude faster than decompressing the image. Can you try this?
```
from PIL import Image
from tqdm import tqdm
from opensfm import dataset


N = 1000

for i in tqdm(range(N)):
    with Image.open('data/berlin/images/01.jpg') as img:
        width, height = img.size

data = dataset.DataSet('data/berlin')
for i in tqdm(range(N)):
    image = data.load_image('01.jpg')
    height, width, _ = image.shape
```

Here Pillow takes less than 1 second while loading the image takes about 3 mins

pierotofy:
Mystery solved, for a test dataset I used some large (like, 15000x15000) images and pillow did not handle them, (`PIL.Image.DecompressionBombError: Image size (224980000 pixels) exceeds limit of 178956970 pixels, could be decompression bomb DOS attack.`) so it went to the (slow) fallback routine (but I didn't notice).

Normal datasets will not have these image sizes.

Thank you for the rigorous review! I've added Pillow to the `requirements.txt` and changed the code to use it.

opensfm/commands/export_visualsfm.py
```diff
@@ -57,6 +57,14 @@ def image_path(self, image, data):
...
-        """Height and width of the undistorted image."""
-        image = data.load_undistorted_image(image)
-        return image.shape[:2]
+        """
+        Height and width of the undistorted image.
+        """
+        try:
+            file_path = data._undistorted_image_file(image)
+            with Image.open(file_path) as img:
+                return img.size
```
paulinus:
```suggestion
                width, height = img.size
                return height, width
```

PIL seems to return `width, height` while we are returning `height, width`

pierotofy:
Ah good catch on the width,height swap :+1: 

I think I'll stop trying to send patches after 8pm from now on. lol, sorry I had to make you review this multiple times.

paulinus:
np, thanks for accelerating this!

PR #425, opened 2019-06-01
Add self parameter in  get_mm_per_unit function

sabado:
When Focal Plane Resolution Unit is present in exif,  function extract_sensor_width calls self.get_mm_per_unit(resolution_unit) with only one argument.  Added self as first argument. 

File "/home/test/OpenSfM/opensfm/exif.py", line 197, in extract_sensor_width
    mm_per_unit = self.get_mm_per_unit(resolution_unit)
TypeError: get_mm_per_unit() takes exactly 1 argument (2 given)

### DIFF
--- a/opensfm/exif.py
+++ b/opensfm/exif.py
@@ -202,7 +202,7 @@ class EXIF:
         width_in_pixels = self.extract_image_size()[0]
         return width_in_pixels * units_per_pixel * mm_per_unit
 
-    def get_mm_per_unit(resolution_unit):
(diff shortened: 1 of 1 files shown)


sabado:
I'm not a Python programmer, and i know this patch may be not the most elegant solution, but works :)

paulinus:
Thanks, @sabado !

sabado:
IMG_4143

@paulinus there is an image with FocalPlaneResolutionUnit tag on it! And thanks you for make this software exists :)

PR #426, opened 2019-06-03
Refactor ground control point constraints

paulinus:
This PR adds a JSON format to specify ground control points, which is easier to use for cases where the altitude and/or location of the point are unknown.

When bundling GCPs are added as regular points to reduce code redundancy.

Future extensions will be
- handling GCPs with unknown location
- allowing providing the accuracy of the measurement.


opensfm/src/bundle/src/projection_errors.h
```diff
@@ -355,122 +353,3 @@ struct BrownInternalParametersPriorError {
   double k3_estimate_;
   double k3_scale_;
 };
-
-struct RotationPriorError {
```
YanNoun: Why does this disappear ?

paulinus: just moved to another file

PR #427, opened 2019-06-03
feat: inliers graph

YanNoun:
This PR robustify the incremental SfM pipeline by explicitely keeping track and storing inliers through the reconstruction process. This result in major improvement on traditional SfM datasets (ETH3D and Strecha). To do so, we had to :

 - Use a second graph (called `inliers_graph`)through the incremental SfM, that keeps track of inliers.
 - We now perform outliers filtering at the individual projection level instead of entire tracks.
 - We introduce a triangulation variant based on robust/RANSAC triangulation. Not used by default as it didn't show any major improvment. See `triangulation_type` in `config`
 - We also introduce a new scheme (`AUTO`) for filtering outliers based on the error distribution rather than a fixed threshold.  Can greatly improve results by is let deactivated for keeping legacy behaviour (`FIXED`). See `bundle_outlier_filtering_type` in `config`.
 - We also remove the penalty-on-behind trick in the projection errors as it is prone to stall the optimizer.

### DIFF
--- a/opensfm/reconstruction.py
+++ b/opensfm/reconstruction.py
@@ -991,7 +991,10 @@ def retriangulate(graph, graph_inliers, reconstruction, config):
         if image in graph:
             tracks.update(graph[image].keys())
     for track in tracks:
-        triangulator.triangulate_robust(track, threshold, min_ray_angle)
+        if config['triangulation_type'] == 'ROBUST':
+            triangulator.triangulate_robust(track, threshold, min_ray_angle)
+        elif config['triangulation_type'] == 'FULL':
+            triangulator.triangulate(track, threshold, min_ray_angle)
 
     report['num_points_after'] = len(reconstruction.points)
     chrono.lap('retriangulate')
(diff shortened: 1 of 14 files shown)


opensfm/reconstruction.py
```diff
@@ -317,7 +318,7 @@ def bundle_local(graph, reconstruction, gcp, central_shot_id, config):
         point = reconstruction.points[point]
         p = ba.get_point(point.id)
         point.coordinates = [p.p[0], p.p[1], p.p[2]]
-        point.reprojection_error = p.reprojection_error
+        point.reprojection_errors = p.reprojection_errors
```
paulinus: The Point type has a `reprojection_error` member that is stored and loaded from JSON files. The new `reprojection_errors` is not stored as of now.

It is convenient not to store the `reprojection_errors` dict since it is going to be large. In this case, we can remove the code to store the old `reprojection_error`.

YanNoun: So we could either store the max of the dict of now ? And set to empty on load ?

paulinus: I think we can skip storing anything since we are not loading it

PR #428, opened 2019-06-05
feat(docs): use brand colors in docs

tofuness:
Some style overrides using brand colors. Doesn't touch `sphinx_rtd_theme`.

WIP.


paulinus:
@tofuness, can we keep a mono spaced font for code blocks?
There are some ascii diagrams and code that requires that. For example, https://www.opensfm.org/docs/geometry.html

tofuness:
@paulinus fixed 🕵

PR #429, opened 2019-06-06
feature: bag-of-words matching

YanNoun:
This PR add bags-of-words (BoW) matching to OpenSfM. This branch is the combination of https://github.com/mapillary/OpenSfM/tree/feat-match-words, which adds BoW matching, and other commits that add pairs generation from BoW. More specifically, we did the following :

 - Compute words assignemnts at feature extraction stage, when either matching or pairs generation is BoW-based.
 - Add image-to-image BoW matching. 2x to 5x faster than FLANN-based matching with reduction in accuracy.
 - Removed unused preemptive matching
 - Moved pair generation code to a new module
 - Added pairs generation using BoW distances, further controlled by `matching_bow_neighbors` in `config`

### DIFF
--- a/opensfm/commands/detect_features.py
+++ b/opensfm/commands/detect_features.py
@@ -90,6 +90,7 @@ def detect(args):
         n_closest = data.config['bow_words_to_match']
         closest_words = bows.map_to_words(
             f_sorted, n_closest, data.config['bow_matcher_type'])
+        closest_words = closest_words[order, :]
         data.save_words(image, closest_words)
 
     end = timer()
--- a/opensfm/commands/detect_features.py
+++ b/opensfm/commands/detect_features.py
@@ -85,7 +85,7 @@ def detect(args):
     if data.config['matcher_type'] == 'FLANN':
         index = features.build_flann_index(f_sorted, data.config)
         data.save_feature_index(image, index)
-    elif data.config['matcher_type'] == 'WORDS':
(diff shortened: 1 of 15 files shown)


opensfm/commands/detect_features.py
```diff
@@ -51,42 +52,51 @@ def write_report(self, data, wall_time):
...
-
-        end = timer()
-        report = {
-            "image": image,
-            "num_features": len(p_sorted),
-            "wall_time": end - start,
-        }
-        data.save_report(io.json_dumps(report),
-                         'features/{}.json'.format(image))
+    start = timer()
```
paulinus: We should skip computing the features if they are already computed.
Before we were using `data.feature_index_exists(image)` that will not work for words so we need something else.

opensfm/src/matching.cc
```diff
@@ -0,0 +1,99 @@
...
+  for (unsigned int i = 0; i < w1.rows; ++i) {
+    int checks = 0;
+    for (unsigned int j = 0; j < w1.cols; ++j) {
+      int word = w1.at<int>(i, j);
+      auto range = index2.equal_range(word);
+      for (auto it = range.first; it != range.second; ++it) {
+        int match = it->second;
+        const float *pa = f1.ptr<float>(i);
+        const float *pb = f2.ptr<float>(match);
+        float distance = DistanceL1(pa, pb, f1.cols);
```
paulinus: Should we put L2 to make it the same distance as the other matchers?

PR #433, opened 2019-06-18
Faster extract_metadata when using use_exif_size: no

pierotofy:
While `extract_metadata` seems I/O bound, I've noticed that processing with two processes yields almost 30% speed-up in performance.

With 78 images:

```
real	0m19.920s
user	0m16.316s
sys	0m3.808s
```

vs. proposed

```
real	0m14.724s
user	0m22.400s
sys	0m6.029s
```

Not sure if the added complexity is worth it, but thought I'd open a PR to see if there's interest.

### DIFF
--- a/opensfm/commands/extract_metadata.py
+++ b/opensfm/commands/extract_metadata.py
@@ -38,12 +38,7 @@ class Command:
 
                 data.save_exif(image, d)
 
-        processes = min(data.config['processes'], 2) # Don't use all CPUs as we're I/O bound
-        if processes > 1:
-            parallel_map(extract_exif, data.images(), processes)
-        else:
-            for image in data.images():
-                extract_exif(image)
+        parallel_map(extract_exif, data.images(), data.config['processes'])
 
         camera_models = {}
         for image in data.images():
@@ -76,7 +71,7 @@ class Command:
 
         # Image Height and Image Width
         if d['width'] <= 0 or not data.config['use_exif_size']:
-            d['height'], d['width'] = data.load_image(image).shape[:2]
+            d['height'], d['width'] = data.image_size(image)
 
         d['camera'] = exif.camera_id(d)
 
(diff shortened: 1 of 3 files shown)


opensfm/commands/extract_metadata.py
```diff
@@ -25,20 +27,29 @@ def run(self, args):
...
+                logger.info('Extracting EXIF for {}'.format(image))
                 d = self._extract_exif(image, data)
 
                 if image in exif_overrides:
                     d.update(exif_overrides[image])
 
                 data.save_exif(image, d)
 
+        processes = min(data.config['processes'], 2) # Don't use all CPUs as we're I/O bound
+        if processes > 1:
```
paulinus: parallel_map will already check if processes is 1

opensfm/commands/extract_metadata.py
```diff
@@ -25,20 +27,29 @@ def run(self, args):
...
+                log.setup()
+                logger.info('Extracting EXIF for {}'.format(image))
                 d = self._extract_exif(image, data)
 
                 if image in exif_overrides:
                     d.update(exif_overrides[image])
 
                 data.save_exif(image, d)
 
+        processes = min(data.config['processes'], 2) # Don't use all CPUs as we're I/O bound
```
paulinus: I'm guessing it is i/o bound only if it needs to decode the entire image

pierotofy: I think you are right.

pierotofy:
Thanks @paulinus! 

After changing the image size calculation to use the same approach as the VisualSFM export code I realized that there's no need for parallelism (the code works faster on a single process when using the optimization), but it's a huge speed-up when `use_exif_size` is set to `no`.

I renamed the PR to reflect this.

Please let me know if further changes are needed.

paulinus:
Looks good. Thanks, @pierotofy!

PR #436, opened 2019-06-22
Export points option for VisualSFM exporter

pierotofy:
This PR adds support to export points along with camera positions in the NVM file format. This can be useful for examination of the output in programs such as VisualSFM or as an input to MVS programs that might need the sparse point cloud as an initialization step (for example https://github.com/alexlocher/hpmvs).

The default behavior of `export_visualsfm` is unchanged, but the `--points` flag now optionally allows to export points along with cameras.

Please let me know if changes are needed :pray:

### DIFF
--- a/opensfm/commands/export_visualsfm.py
+++ b/opensfm/commands/export_visualsfm.py
@@ -6,6 +6,7 @@ import os
 from opensfm import dataset
 from opensfm import transformations as tf
 from opensfm import io
+from opensfm import types
 from six import iteritems
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


paulinus:
Looks good to me. Thanks, @pierotofy !

About exporting the Brown camera model, one thing to keep in mind is that VisualSfM camera model is different and we are not even exporting the radial distortion parameters.

So the only way to make sure that what we export is properly read by VisualSfM is to export the undistorted reconstruction (using the `--undistorted`) option. This way all cameras are perspective with no distortion.

pierotofy:
Thanks for the review @paulinus !

PR #439, opened 2019-06-26
Merge OpenSfM and production matching 1/2

YanNoun:
This PR is the first in a series of 2, which aims at unifiying OpenSfM matching with Mapillary's production matching. This PR focuses on the pair selection part, and adds the following :

 - Preempt images using GPS before doing BoW matching, This an be controlled using `matching_bow_gps_distance` and `matching_bow_gps_neighbors`.
 - Add an option for enforcing matching between different camera using `matching_bow_other_cameras`
 - Switch parallel tasking back-end to `joblib`, replacing `loky`.
 - Introduce the use of two sets when doing matching `reference` and 'candidates`. By default, these are the same.


opensfm/pairs_selection.py
```diff
@@ -81,70 +31,158 @@ def match_candidates_by_distance(images, exifs, reference, max_neighbors, max_di
         return set()
     max_neighbors = max_neighbors or 99999999
     max_distance = max_distance or 99999999.
-    k = min(len(images), max_neighbors + 1)
+    k = min(len(images_cand), max_neighbors)
+
+    if(images_ref == images_cand):
+        k = k + 1
```
paulinus: It can happen that `image_ref` intersects with `images_cand` and we still want `k+1` candidates since the first one will be the main image.
Maybe we can always use `k+1` and deal with the exact number of candidates later when we build the list of pairs.

opensfm/pairs_selection.py
```diff
@@ -81,70 +31,158 @@ def match_candidates_by_distance(images, exifs, reference, max_neighbors, max_di
...
     """Find candidate matching pairs by time difference."""
     if max_neighbors <= 0:
         return set()
-    k = min(len(images), max_neighbors + 1)
+    k = min(len(images_cand), max_neighbors)
 
-    times = np.zeros((len(images), 1))
-    for i, image in enumerate(images):
+    if(images_ref == images_cand):
+        k = k + 1
```
paulinus: same here

PR #443, opened 2019-07-11
Feat gcp without coordinates

paulinus:
This PR makes it possible to provide Ground Control Points for which the world position is unknown. That is point correspondences that are annotated in the images but for which the world lat, lon position has not been measured.


bin/plot_gcp.py
```diff
@@ -67,19 +71,25 @@ def main():
...
+    data = dataset.DataSet(args.dataset)
     reconstruction = data.load_reconstruction()[0]
     gcps = data.load_ground_control_points()
 
     with io.open_wt(data.data_path + '/gcp.ply') as fout:
-        fout.write(gcp_to_ply(gcps))
+        fout.write(gcp_to_ply(gcps, reconstruction))
 
     for gcp in gcps:
+        print("Plotting GCP '{}'".format(gcp.id))
```
YanNoun: `logger.info` ?

paulinus: 👍 i'll put it on the figure's tittle

Issue #413, opened 2019-04-08
[Question] GCP alignment


pierotofy:
Hello :hand:

I have a small dataset captured from a drone. Each image has GPS lat/lon/alt information included.

For additional accuracy I've tried to include a GCP file. The resulting point cloud however, exported with `opensfm export_geocoords --dense` does not align well with the GCPs (it's offset by ~1 meter in the X and ~3 meters in the Y axix).

Point cloud coordinates (Xg and Yg) of GCP point (WGS84 UTM 16N):

![image](https://user-images.githubusercontent.com/1951843/55747733-eafaf980-5a0a-11e9-9f18-0706f4461f30.png)

Entry from GCP file:

```
609865.7259757777 4950688.244601672 170.675678 2704 2559 DJI_0027.JPG
```

GCP point (top left) vs actual position.

![image](https://user-images.githubusercontent.com/1951843/55747829-2a294a80-5a0b-11e9-94eb-d0433159f886.png)

OpenSfM config file:

```
use_exif_size: no
optimize_camera_parameters: yes
use_altitude_tag: True
align_method: naive
bundle_use_gcp: yes
bundle_use_gps: no
```

I'm wondering if this is expected, if I need to toggle a setting, or if there's something I should check (or did I find a bug)?

I should mention there are 5 GCPs in total and none of them align correctly.

YanNoun:
Hi,

I'm not sure what went wrong, but the way we do align is by realigning the reconstruction on GPS+GCP after any whole bundle adjustment. It might not be the optimal way of doing it, and we're currently working on it.

So at the very end of the process, after last bundle, the reconstruction is aligned on GPS+GCP. It's already misleading since you might want not to use GPS (as GPS can be off by several meters compared to GCPs).

I made a tentative-untested branch `feat-gcp-align` that does not align on GPS AT ALL is `bundle_use_gps` is no, to see if it solves your case. let me know how it goes.

Yann

pierotofy:
Hey @YanNoun, thanks for the reply! :hand:

Like you say, it looks like both GPS+GCP observations are used even though `budle_use_gcp` is set to `no`.

I will try the changes in your branch and report back my findings.

pierotofy:
Looking good!!

![image](https://user-images.githubusercontent.com/1951843/59802982-2c840c80-92b8-11e9-9acd-b3de48131df1.png)

[...]

PR #449, opened 2019-07-26
feat: better gps and gcp align

YanNoun:
This PR aims at improving the alignment behaviour when using GPS and GCP.

More specifically, we propagate the use of `bundle_use_gps` and `bundle_use_gcp` in the alignment. Discussion is welcome whether we should have a distinct option or not.

The global behaviour is also changed as it is done before any bundle, hoping GPS/GCP error will be closer to optimum when starting the bundle. Additionally, we also support edge cases when we have only one or two GCP/GPS (scale/orientation ambiguity will remain though, no a-priori is being made).

Finally, we refactored code a bit so we can support correct topocentric frame when there is no GPS.

Yann

### DIFF
--- a/opensfm/align.py
+++ b/opensfm/align.py
@@ -58,8 +58,7 @@ def align_reconstruction_similarity(reconstruction, gcp, config):
         return align_reconstruction_naive_similarity(config, reconstruction, gcp)
 
 
-def align_reconstruction_naive_similarity(config, reconstruction, gcp):
-    """Align with GPS and GCP data using direct 3D-3D matches."""
+def alignment_constraints(config, reconstruction, gcp):
     X, Xp = [], []
 
     # Get Ground Control Point correspondences
@@ -73,6 +72,13 @@ def align_reconstruction_naive_similarity(config, reconstruction, gcp):
         for shot in reconstruction.shots.values():
             X.append(shot.pose.get_origin())
             Xp.append(shot.metadata.gps_position)
+
+    return X, Xp
+
+
+def align_reconstruction_naive_similarity(config, reconstruction, gcp):
+    """Align with GPS and GCP data using direct 3D-3D matches."""
+    X, Xp = alignment_constraints(config, reconstruction, gcp)
 
     if len(X) == 0:
         return 1.0, np.identity(3), np.zeros((3))
@@ -114,12 +120,10 @@ def align_reconstruction_orientation_prior_similarity(reconstruction, config):
      - horizontal: assumes cameras are looking towards the horizon
      - vertical: assumes cameras are looking down towards the ground
     """
-    X, Xp = [], []
+    X, Xp = alignment_constraints(config, reconstruction, gcp)
     orientation_type = config['align_orientation_prior']
     onplane, verticals = [], []
     for shot in reconstruction.shots.values():
-        X.append(shot.pose.get_origin())
-        Xp.append(shot.metadata.gps_position)
         R = shot.pose.get_rotation_matrix()
         x, y, z = get_horizontal_and_vertical_directions(
             R, shot.metadata.orientation)
--- a/opensfm/align.py
+++ b/opensfm/align.py
@@ -88,7 +88,7 @@ def detect_alignment_constraints(config, reconstruction, gcp):
 
     X, Xp = alignment_constraints(config, reconstruction, gcp)
     if len(X) < 3:
-        return 'naive'
+        return 'orientation_prior'
 
     X = np.array(X)
     X = X - np.average(X, axis=0)
--- a/opensfm/align.py
+++ b/opensfm/align.py
@@ -55,26 +55,38 @@ def align_reconstruction_similarity(reconstruction, gcp, config):
         return align_reconstruction_orientation_prior_similarity(
             reconstruction, config)
     elif align_method == 'naive':
-        return align_reconstruction_naive_similarity(reconstruction, gcp)
+        return align_reconstruction_naive_similarity(config, reconstruction, gcp)
 
 
-def align_reconstruction_naive_similarity(reconstruction, gcp):
+def align_reconstruction_naive_similarity(config, reconstruction, gcp):
     """Align with GPS and GCP data using direct 3D-3D matches."""
     X, Xp = [], []
 
     # Get Ground Control Point correspondences
-    if gcp:
+    if gcp and config['bundle_use_gcp']:
         triangulated, measured = triangulate_all_gcp(reconstruction, gcp)
         X.extend(triangulated)
         Xp.extend(measured)
 
     # Get camera center correspondences
-    for shot in reconstruction.shots.values():
-        X.append(shot.pose.get_origin())
-        Xp.append(shot.metadata.gps_position)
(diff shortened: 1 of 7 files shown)


YanNoun:
Let's add orientation prior in the alignment.

opensfm/align.py
```diff
@@ -55,26 +55,38 @@ def align_reconstruction_similarity(reconstruction, gcp, config):
...
+        return 1.0, np.identity(3), np.zeros((3))
+
+    # Translation-only case
+    if len(X) == 1:
+        t = Xp[0] - X[0]
+        return 1.0, np.identity(3), t
+
+    # For now, translation only, need to decide on a prior rotation
+    if len(X) == 2:
+        s = np.linalg.norm(Xp[0] - Xp[1])/np.linalg.norm(np.array(X[0]) - np.array(X[1]))
```
pierotofy: This should be:

`s = np.linalg.norm(np.array(Xp[0]) - np.array(Xp[1]))/np.linalg.norm(np.array(X[0]) - np.array(X[1]))`

Per the problem that was discussed in #413 (see comment from Jun 20).

pierotofy:
Finally had time to test this (sorry it took so long). @YanNoun 

Running this branch with these input images + GCP https://github.com/OpenDroneMap/odm_data_copr still produces a reconstruction that is out of scale:

image

The green/red/blue dots are the cameras exported with `export_ply` (so the reconstruction looks really tiny). I'm going to investigate this a bit further to see if I can find a cause/fix.

pierotofy:
Using:

```
align_method: naive
```

Scale looks OK, but rotation/translation does not.

image

image

pierotofy:
Interestingly, the GCP locations seem to be triangulated fine.

image

I've extracted the GCP coordinates by adding a `print` in `_add_gcp_to_bundle`:

`reconstruction.py`
...
        coordinates = triangulate_gcp(point, shots)
        if coordinates is None:
            if point.coordinates is not None:
                print(point.coordinates)
                coordinates = point.coordinates
            else:
                logger.warning("Cannot initialize GCP '{}'."
                               "  Ignoring it".format(point.id))
                continue
...

```
[-4.85703576e+00  4.65463388e+00 -3.55299562e-06]
[ 1.44868850e+01 -1.80340258e+01 -4.20194119e-05]
[-9.43289669e-01 -5.33349274e+00 -2.30874866e-06]
[-1.61153324e+01  1.79837399e+01 -4.57800925e-05]
[ 6.54344838e+00 -1.05103216e+01 -1.20438635e-05]
[ 7.15283159e-01  4.91304974e+00 -1.93994492e-06]
[-1.77048913e+01  1.12889474e+01 -3.45725566e-05]
[ 6.40999243e+00 -5.46441798e+00 -5.56744635e-06]
[-6.28309068e+00  1.36635485e+01 -1.77789479e-05]
[ 1.77481583e+01 -1.31615936e+01 -3.82959843e-05]
2019-08-13 20:17:52,976 DEBUG: Ceres Solver Report: Iterations: 4, Initial cost: 1.261175e+02, Final cost: 1.245557e+02, Termination: CONVERGENCE
2019-08-13 20:17:53,140 INFO: Removed outliers: 1
2019-08-13 20:17:53,169 INFO: {'average_track_length': 2.3795516546079942, 'points_count': 8431, 'observations_count': 20062, 'cameras_count': 3, 'average_track_length_notwo': 3.0}
2019-08-13 20:17:53,169 INFO: Reconstruction 0: 38 images, 80350 points
2019-08-13 20:17:53,169 INFO: Reconstruction 1: 3 images, 8431 points
2019-08-13 20:17:53,169 INFO: 2 partial reconstructions in total.
```

pierotofy:
(off-topic comment, I've got BOW matching to work, it's AWESOME! )

pierotofy:
Can't seem to find anything too obvious. But seems like the last call to `bundle` fails to converge to a solution:

```
2019-08-14 02:50:09,600 DEBUG: Ceres Solver Report: Iterations: 51, Initial cost: 1.318168e+05, Final cost: 5.121470e+04, Termination: NO_CONVERGENCE
2019-08-14 02:50:15,185 INFO: Removed outliers: 785
2019-08-14 02:50:15,441 INFO: {'average_track_length': 3.75419755560008, 'points_count': 74865, 'observations_count': 281058, 'cameras_count': 38, 'average_track_length_notwo': 4.869867354297328}
```

Please let me know if I can provide additional information / test data. :pray:

YanNoun:
Hi @pierotofy ,

Spend some time investigating what's going on. First some observations :
 - First of all, there is only 2 GCPs having at least 2 projections, thus being triangulated (all the others GCP have one projection). We recommend at least 3 GCPs with >= 2 projections (let's call them triangulated GCPs) for proper alignment.
 - Thanks to the `orientation_prior`, OpenSfM can align even with 1 or 2 triangulated GCPs.

Then, what I fixed :
 - There was an obvious bug (not solving for the rotation) in the 2 triangulated GCPs case. It is solved now.

Then, the most problematic point, the data :
 - You can use the command `./bin/plot_gcp` to see the GCP projected (blue) in the reconstructed image versus its annotated position in the image (orange). In our case, the alignment of the two triangulated GCPs will be perfect (exact solution, orange and blue superposed) but you will see some huge discrepancies on the others GCPs (1 projections ones), though, which leads me to the next point.
 - I never managed to get any consistent alignment with 3 or 4 triangulated GCPs (I added some more projections). I processed the project with Agisoft/Metashape as reference with the same GCP data and I had error of 1 meter. Both with `WGS84` and `UTM` coordinates. See below (`source` mean GCP position, other being the estimated position). So I wonder what's wrong with the data itself (what's the accuracy of the surveyed data ?).
agisoft_odm
 - I had the same magnitude of 3D error (1 meter) in OpenSfM and Agisoft. Somehow a good news.

I finally added a feature to not bother anymore with alignment method :
 - I added a `auto` mode as `align_method` which will automatically switch between `naive` and `orientation_prior` depending on the GCP configuration (if they lie on a line or not).

Would be good if we could test on some other dataset with better GCPs.

Yann

pierotofy:
Thanks for looking into this @YanNoun ! Helpful as always.

I think your observations are correct, I don't think these measurements were collected with precision equipment. @smathermather might know the background story for how this data was collected, but I think it's safe to assume an error of a few meters, which is within the results obtained. Like you mentioned, only two GCPs are tied to two images, which is not ideal.

With the latest changes and using `auto` (nice!) the result looks at scale and in the right place:

image

:tada: 

I will test two more datasets that I know have better accuracy and post my results here.

pierotofy:
Looks good on a dataset with both GPS and GCPs.

image

pierotofy:
And here's another one:

image

image

image

image

image

pierotofy:
Would be helpful, yes.

denistestemale:
Hello.
Sorry for commenting so late. One of the dataset was mine, obtained with <2cm precision. With a recent WebODM version (updated april 28th 2020) the results are very good indeed:
- precision is now below 8cm for all GCPs, on several datasets obtained with the same GPS;
- doming effect disappeared (might be related). 

Thanks!
d

PR #459, opened 2019-08-15
Added camera_projection_type config

pierotofy:
Hello :hand:

This PR proposes the addition of a new `camera_projection_type` config to forcibly set the camera model, ignoring the information from EXIFs. This only works for sets of images that all share the same camera model, but existing behavior remains unchanged by leaving the default set to `AUTO`.

We've found that many cameras lack the metadata necessary to be detected as fisheye (the XMP tag `GPano:ProjectionType` is lacking in many cameras) and it seems that `perspective` is (always?) favored over `brown`.

I think I've interpreted the default values for `c_x` and `c_y` correctly (0,0 represents the center?) and I've added a fallback while retrieving the `focal_x`/`focal_y` keys to use `focal_ratio`.

I might have missed something while implementing this, in which case please let me know and I'll make adjustments. :+1:

If this seems like a terrible idea, also let me know. xD

### DIFF
--- a/opensfm/commands/extract_metadata.py
+++ b/opensfm/commands/extract_metadata.py
@@ -63,6 +63,10 @@ class Command:
          # EXIF data in Image
         d = exif.extract_exif_from_file(data.open_image_file(image))
 
+        # Override projection type if needed
+        if data.config['camera_projection_type'] != 'AUTO':
+            d['projection_type'] = data.config['camera_projection_type'].lower()
+
         # Image Height and Image Width
         if d['width'] <= 0 or not data.config['use_exif_size']:
             d['height'], d['width'] = data.image_size(image)
--- a/opensfm/config.py
+++ b/opensfm/config.py
@@ -93,6 +93,7 @@ bundle_outlier_filtering_type: FIXED    # Type of threshold for filtering outlie
 bundle_outlier_auto_ratio: 3.0          # For AUTO filtering type, projections with larger reprojection than ratio-times-mean, are removed
 bundle_outlier_fixed_threshold: 0.006   # For FIXED filtering type, projections with larger reprojection error after bundle adjustment are removed
 optimize_camera_parameters: yes         # Optimize internal camera parameters during bundle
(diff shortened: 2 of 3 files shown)


paulinus:
Hi @pierotofy, there are ways to force OpenSfM to use specific camera parameters but not well documented. Thanks for looking at how to make it simpler.

In general, there might be more than one camera in a dataset.  So OpenSfM has a way to set specific camera parameters for each camera id by creating the `camera_models_overrides.json` file with a dictionary mapping camera ids to camera parameters.

This is a bit tedious to do because we usually do not know the camera id beforehand and we do not know the format we should specify the parameters in. A way to solve this is to run `extract_metadata` once, then rename `camera_models.json` to `camera_models_overrides.json`, and then edit it.

But there is a simpler way of doing this, which was not documented.  You can use `all` as a camera id in `camera_models_overrides.json` and the parameters will be used for all cameras. This way you can create the `camera_models_overrides.json` file without having to run `extract_metadata` first.

I've added some documentation on how this works here https://www.opensfm.org/docs/using.html#providing-your-own-camera-parameters

I think this should be usable for your use case without having to add a new option. What do you think?

pierotofy:
Thanks @paulinus, yes I thought about using the camera overrides, and thought to contribute this patch for the reasons you mentioned (avoiding to parse the file for camera IDs). I did not know about the "all" camera ID, thanks for mentioning it. I should have read the source more thoroughly :)

With the "all" key I don't think this PR is needed. I'll go ahead and close it.

pierotofy:
I should mention, the changes in `exif.py` https://github.com/mapillary/OpenSfM/pull/459/commits/267fdbd07bbf7db1f1815f8d71b2a853f258e12d should still be needed, I can open a separate PR for that if they are.

pierotofy:
Also perhaps the changes in `exif.py` from https://github.com/mapillary/OpenSfM/pull/459/commits/3c74cd40026efcb7b92c0c2b10f3c3d6a982690f

paulinus:
I've cherry-picked those changes, thanks!

PR #461, opened 2019-08-23
feat: render function takes borderMode option

mlopezantequera:


paulinus:
I was thinking that panoramas should not have a border. Are we using that method for images that are not panoramas? Maybe we need a better name for it

mlopezantequera:
@paulinus I am only using it on non-panoramic images to double check my re-implementation of this in torch, but since it does indeed work for any pair of shots we might want to rename it

mlopezantequera:
some ideas:
`re_render_shot`
`render_new_shot`
`render_shot`
`project_and_render`
`reproject_and_render`

paulinus:
ok, will pick another name for it. Merging the border changes already. Thanks.

PR #462, opened 2019-08-25
BROKEN  fix: get the number of features right

paulinus:
There is something funky going on here, @YanNoun.

Before this fix, we are always rebuilding the index because we get `len(current_features)==3`

After the fix, running on lund we get a segmentation fault. The problem is that the flann index is created with a filtered array here
https://github.com/mapillary/OpenSfM/blob/master/opensfm/matching.py#L107-L108
This array is not cached and gets garbage collected. So the second time that we try to use the same index, it points to that array that no longer exists.

One way to solve that would be to filter the feature arrays before caching them.  It could be done in the feature loader itself.

### DIFF
--- a/opensfm/feature_loading.py
+++ b/opensfm/feature_loading.py
@@ -49,7 +49,7 @@ class FeatureLoader(object):
 
     def load_features_index(self, data, image, features):
         index = self.index_cache.get(image)
-        current_features = self.load_points_features_colors(data, image)
+        _, current_features, _ = self.load_points_features_colors(data, image)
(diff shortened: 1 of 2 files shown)


paulinus:
Fixed the feature array missing problem by caching it together with the index

paulinus:
Currently, the tests fail because a test tries to `load_feature_index` but the test dataset has no such method and returns no index.

I suggest not fixing this and instead always build the index instead of loading it as done in #463

YanNoun:
That was a nasty one ... !

PR #466, opened 2019-08-30
VLAD and FLANN on the fly

oscarlorentzon:
This PR adds VLAD as an option for image retrieval and FLANN calculated on the fly (without storing any files) as an option for initial matching.

The VLAD vectors are constructed using a trained vocabulary of 64 words and thereafter [SSR normalized](https://www.robots.ox.ac.uk/~vgg/publications/2013/arandjelovic13/arandjelovic13.pdf).

The VLAD processing and options are equivalent to that of the BOW image retrieval.

The feature loader instance cache needs to be cleared before the unit tests are run because the naming of the lund images are overlapping some of the berlin images. The long running unit tests calling `pairs_selection.match_candidates_from_metadata` could be refactored to avoid having to detect features for each of the tests if the time running the tests is considered too long.


opensfm/test/test_pairs_selection.py
```diff
@@ -0,0 +1,143 @@
...
+    for name in names:
+        if name in matches:
+            count += 1
+
+    assert count >= assert_count
+
+
+def create_match_candidates_config(**kwargs):
+    config={
+        str('matcher_type'): str('BRUTEFORCE'),
```
paulinus: Do we need to convert strings to strings?

oscarlorentzon: This is because of the future import:

`from __future__ import unicode_literals` 

It was present in other test files so I included it here too. If the literals are not converted to strings the written config file becomes corrupt and not loadable.

oscarlorentzon:
Note that the unit test stage seems to take ~6 min more with the addition of the unit tests using the Lund dataset.

opensfm/test/test_pairs_selection.py
paulinus: ok, that means we need to fix the config I/O and keep using Unicode literals everywhere.
I'll have a look and remove the str() conversions when fixed

oscarlorentzon: The config file needs to be written to disc because of direct command calls like so:

```
commands.extract_metadata.Command().run(args)
commands.detect_features.Command().run(args)
```

oscarlorentzon: OK, thanks.

PR #467, opened 2019-08-30
Fix mask for rotated images

paulinus:
We had a bug on the rotation matrix that rotates features according to the exif tag.
This PR fixes the bug and refactors the feature extraction code to always use the same mask.


paulinus:
👍 the `load_feature_mask` is great and makes it easy to fix the problem because it put all rotations and masking in a single place.
The problem was with the rotation matrices, which I got wrong in the original code.

PR #473, opened 2019-09-02
fix: use DJI XMP EXIF and handle same-GPS case

YanNoun:
This PR aims at fixing two GPS issues :
 - Defective GPS values, all values being the same. Which lead to further issue in computing alignment.
 - DJI drones have buggy firmware versions (Mavic), where classic `GPSLatitude` tags had truncated values (leading to same-values issue, see above). Using DJI-specific XMP tags when available mitigates the issue.

Ping @pierotofy if you can test on more datasets. I got the issue on `hellasgarden' dataset taken with a Mavic Air (?), but can't remember where I got it from, thought it was from OpenSfM GH issues.

### DIFF
--- a/opensfm/align.py
+++ b/opensfm/align.py
@@ -116,9 +116,16 @@ def align_reconstruction_naive_similarity(config, reconstruction, gcp):
     if len(X) == 0:
         return 1.0, np.identity(3), np.zeros((3))
 
-    # Translation-only case
-    if len(X) == 1:
+    # Translation-only case, either : 
+    #  - a single value
+    #  - identical values
+    same_values = (np.linalg.norm(np.std(Xp, axis=0)) < 1e-10)
+    single_value = (len(X) == 1)
+    if single_value:
         logger.warning('Only 1 constraints. Using translation-only alignment.')
+    if same_values:
+        logger.warning('GPS/GCP data seems to have identical values. Using translation-only alignment.')
+    if same_values or single_value:
         t = np.array(Xp[0]) - np.array(X[0])
         return 1.0, np.identity(3), t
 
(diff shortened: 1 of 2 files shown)


opensfm/align.py
```diff
@@ -128,6 +128,13 @@ def align_reconstruction_naive_similarity(config, reconstruction, gcp):
         X.append(X[1])
         Xp.append(Xp[1])
 
+    # Sometimes, consraint data such as GPS EXIFs can
+    # be corrupted, all values being the same
+    std = np.std(Xp, axis=0)
+    if np.linalg.norm(std < 1e-10):
+        logger.warning('GPS/GCP data seems to have identical values : unable to align.')
+        return 1.0, np.identity(3), np.zeros((3))
```
paulinus: Should we handle this case in the same way we handle the `len(X) == 1` above?
So no rotation or scale is estimated but only translation.
Maybe we can join the two conditions in a single if clause.

YanNoun: Done.

pierotofy:
Happy to help test this @YanNoun when I come back from vacation after September 9th. The hellasgarden dataset was also posted on https://community.opendronemap.org/t/weird-result-with-a-mavic-air/2722 and is available from https://drive.google.com/file/d/13tPpJMgmEEq2G04KWxTFRx6RzNUxcb5m/view

I don't personally have datasets collected with a mavic air, if I had to test it I would have also used the hellasgarden dataset. Didn't find other datasets on the opendronemap forum.

pierotofy:
* XMP values are extracted correctly (verified by looking at `exif/` folder)
 * Alignment looks good (what I would expect to see from a DJI GPS, see below)

image

anim

 * Reconstruction exhibits doming (see below), but I suspect it's because of the flight path / camera:

image

image

Looks good to me! :+1:

YanNoun:
Merged. Added the `AbsoluteAltitude` retrieval.

PR #475, opened 2019-09-02
Update reconstruction.html

Arshiamidos:
add support for point_clouds  
when you toggle `drawLineOnPointClick` and `selectedCamera` is fullfilled, it's draw line between `active Camera` and `point_cloud`

Screen Shot 1398-06-12 at 0 40 34

### DIFF
--- a/viewer/reconstruction.html
+++ b/viewer/reconstruction.html
@@ -277,6 +277,7 @@
             var num_preview_plane = 5;
             var moveSpeed = 0.2;
             var turnSpeed = 0.1;
+            var virtualLinesByClick = new THREE.Object3D();
             var previousShot = undefined;
             var validMoves;
             var movingMode = 'orbit';
@@ -285,6 +286,7 @@
                 pointSize: 0,
                 showThumbnail: false,
                 showImagePlane: true,
+                drawLineOnPointClick:false,
                 drawGrid: false,
                 drawGPS: false
             };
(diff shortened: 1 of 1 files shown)


viewer/reconstruction.html
```diff
@@ -1112,6 +1126,33 @@
...
+
+                intersected = raycaster.intersectObjects(point_clouds);
+                if (
+                    intersected.length > 0 &&
+                    options.drawLineOnPointClick && 
+                    selectedCamera
+                ){
+                    var material = new THREE.LineBasicMaterial( { color: 0xFFC0CB } );
+                    var geometry = new THREE.Geometry();
+                    geometry.vertices.push(new THREE.Vector3(
```
paulinus: This will use the center of the bounding sphere but it should use the optical center of the camera. You can use something like 
```
                    var shot = selectedCamera.reconstruction.shots[selectedCamera.shot_id];
                    var ocenter = opticalCenter(shot);
                    geometry.vertices.push(ocenter);
```
to use the optical center instead

viewer/reconstruction.html
```diff
@@ -1112,6 +1126,33 @@
...
+                    selectedCamera
+                ){
+                    var material = new THREE.LineBasicMaterial( { color: 0xFFC0CB } );
+                    var geometry = new THREE.Geometry();
+                    geometry.vertices.push(new THREE.Vector3( 
+                        selectedCamera.geometry.boundingSphere.center.x,
+                        selectedCamera.geometry.boundingSphere.center.y,
+                        selectedCamera.geometry.boundingSphere.center.z,
+                    ));
+                    geometry.vertices.push(new THREE.Vector3(
```
paulinus: similarly, this is using the intersection point (which depends on the position of the mouse over the square representing the point)  but it would be better to use the coordinates of the point that have been selected

PR #477, opened 2019-09-04
Decrease unit test time

oscarlorentzon:
This PR decreases the time it takes to run the unit test stage in CI by ~120 seconds. The unit test run time is still ~50 seconds longer than before the merge of https://github.com/mapillary/OpenSfM/pull/466 because of the feature extraction of the lund dataset which is required for the `match_candidates_from_metadata` tests.

 With this PR the features are extracted once in a module fixture instead of once for each of the four tests.

### DIFF
--- a/opensfm/test/test_pairs_selection.py
+++ b/opensfm/test/test_pairs_selection.py
@@ -17,11 +17,45 @@ from opensfm import feature_loader
 from opensfm.test import data_generation
 
 
+NEIGHBORS = 6
+
+
+class Args():
+    def __init__(self, dataset):
+        self.dataset = dataset
+
+
 @pytest.fixture(scope="module", autouse=True)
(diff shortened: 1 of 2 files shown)


opensfm/test/test_pairs_selection.py
```diff
@@ -117,27 +143,31 @@ def create_match_candidates_config(**kwargs):
     return config
 
 
-def test_match_candidates_from_metadata_vlad(tmpdir):
+def test_match_candidates_from_metadata_vlad(lund_path):
     config = create_match_candidates_config(matching_vlad_neighbors=NEIGHBORS)
```
paulinus: Should we set `matching_gps_distance: 0` to make sure we don't use gps based neighbors in addition to vlad neighbors?

PR #482, opened 2019-09-25
Add support for more FocalPlaneResolutionUnit values

pierotofy:
Hello! :hand:

It looks like `get_mm_per_unit` was missing a few enum values for units. This causes errors with certain camera vendors (found this from a Parrot Sequoia dataset). The enum values are taken (per  comment indications) directly from https://www.sno.phy.queensu.ca/~phil/exiftool/TagNames/EXIF.html (`FocalPlaneResolutionUnit`).

I've introduced the `cm_in_mm` and `um_in_mm` variables to be consistent with the `inch_in_mm` conversion. 

:clinking_glasses:

### DIFF
--- a/opensfm/exif.py
+++ b/opensfm/exif.py
@@ -14,7 +14,8 @@ from opensfm import types
 logger = logging.getLogger(__name__)
 
 inch_in_mm = 25.4
-
+cm_in_mm = 10
+um_in_mm = 0.001
 
 def eval_frac(value):
     try:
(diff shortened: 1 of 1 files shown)


GITNE:
We had this discussion before. The problem is that these are outside of the Exif specification, only inch and cm have been specified. Cameras which use other values (units) than those specified violate the Exif specification. If you still want to use images from such cameras with OpenSfM, you are basically left with either recomputing and updating the affected values or you run a privately patched version of OpenSfM on your machine(s).

GITNE:
@YanNoun Please revert this PR. These units are **NOT** standard complient. This PR will eventually pollute databases with bad data. Please check with @paulinus.

paulinus:
Thanks for the warning, @GITNE. 

On the one hand, I understand the concern about supporting unofficial standards, as we discussed here https://github.com/mapillary/OpenSfM/commit/a074feab193776569ff23ce0ff5cab31535e34ff#r31973708

On the other hand, while not official, the `mm` and `um` units seem to be used by manufacturers currently and it is good that OpenSfM's supports it. We do not run this particular exif extraction code for the images in the Mapillary database and we are moving to rely the less and less on the exif focal length data. So this change will not affect Mapillary.

So, all in all, I think we are good with the change that will make more cameras work.

PR #485, opened 2019-10-08
Log projection types for matching pairs

oscarlorentzon:


opensfm/matching.py
```diff
@@ -79,6 +83,29 @@ def match_images_with_pairs(data, exifs, ref_images, pairs):
     return resulting_pairs
 
 
+def log_projection_types(pairs, exifs):
+    projection_type_pairs = {}
+    for im1, im2 in pairs:
+        pt1 = exifs[im1]['projection_type']
```
paulinus: If we want to log the actual projection type used, we should look at the camera that corresponds to the camera id in the exif as in here
https://github.com/mapillary/OpenSfM/blob/62fd27b137a348203ed40d9362191dfcbb0a4936/opensfm/matching.py#L116

The projection type in exif may differ or not exist.

oscarlorentzon: Thanks, fixed.

PR #490, opened 2019-10-21
Preserve 16bit information when undistorting 16bit TIFFs

pierotofy:
Hello :hand:

This is a draft PR to preserve 16bit information when undistorting images (such as some TIFF and PNG files). Don't merge yet, there are things to be addressed.

Adding `IMREAD_ANYDEPTH` to `imread` allows the images to be undistorted without losing the 16bit information.

Two problems to solve:

1) The sparse point cloud's colors are off (probably because they are out of range / scaled improperly):

image

2) The dense reconstruction has significant noise:

image

I think I will make changes to use the `IMREAD_ANYDEPTH` flag only when invoking the undistort command.


pierotofy:
The anydepth flag is now used only when undistorting images, preserving 16bit information, but leaving the rest of the pipeline unchanged.

Dense/sparse point clouds reconstruct as expected.

image

pierotofy:
I think this can be ready for review.

paulinus:
Looks good, @pierotofy. Thanks for looking at that.
The tests are failing probably because of changes in the travis build system in master. Could you rebase to master?

pierotofy:
Done! Thanks for the review @paulinus.

PR #494, opened 2019-11-15
feat: added dual camera model

YanNoun:
This PR introduce a new camera model dubbed `dual` than can seamlessly switch between fisheye and perspective projection using an additional parameter called `transition`. More formally, one can rewrite the projection as : 

`proj(dual) = transition * proj(perspective) + (1 - transition) * proj(fisheye)`

Inverse projection (aka. `bearings`) is done using an inverse iterative scheme (only 3 iterations are sufficient), as one can't retrieve the incident angle `theta` from the the distorted points.

`transition` is soft-constrained between 0 and 1 in the bundle, using a log-barrier penalty.

Corresponding `mapillary_sfm` commit is here : https://github.com/mapillary/mapillary_sfm/commit/3330e863714877e8c6674a4e28adbd34b68b3607

### DIFF
--- a/opensfm/exif.py
+++ b/opensfm/exif.py
@@ -482,9 +482,6 @@ def camera_from_exif_metadata(metadata, data):
         camera.focal = calib['focal']
         camera.k1 = calib['k1']
         camera.k2 = calib['k2']
-        camera.focal_prior = calib['focal']
-        camera.k1_prior = calib['k1']
-        camera.k2_prior = calib['k2']
         camera.transition = calib['transition']
         return camera
     elif pt in ['equirectangular', 'spherical']:
--- a/opensfm/exif.py
+++ b/opensfm/exif.py
@@ -470,6 +470,23 @@ def camera_from_exif_metadata(metadata, data):
         camera.k1 = calib['k1']
         camera.k2 = calib['k2']
         return camera
+    elif pt == 'dual':
+        calib = (hard_coded_calibration(metadata)
+                 or focal_ratio_calibration(metadata)
+                 or default_calibration(data))
+        camera = types.DualCamera()
+        camera.id = metadata['camera']
+        camera.width = metadata['width']
+        camera.height = metadata['height']
+        camera.projection_type = pt
+        camera.focal = calib['focal']
+        camera.k1 = calib['k1']
+        camera.k2 = calib['k2']
+        camera.focal_prior = calib['focal']
+        camera.k1_prior = calib['k1']
+        camera.k2_prior = calib['k2']
+        camera.transition = calib['transition']
+        return camera
     elif pt in ['equirectangular', 'spherical']:
         camera = types.SphericalCamera()
         camera.id = metadata['camera']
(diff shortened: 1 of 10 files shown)


opensfm/mesh.py
```diff
@@ -20,12 +20,23 @@ def triangle_mesh(shot_id, r, graph, data):
...
     elif shot.camera.projection_type in ['equirectangular', 'spherical']:
         return triangle_mesh_equirectangular(shot_id, r, graph)
     else:
         raise NotImplementedError
 
 
+def triangle_mesh_dual(shot_id, r, graph):
+    shot = r.shots[shot_id]
+    cam = shot.camera
+    if cam.transition < 0.5:
```
paulinus: There is probably no reasonable threshold to decide this. The fisheye method will work for all camera types as long as their field of view is smaller than 180. So we could always use the fisheye method for simplicity. This way we will at least easily know which method was used.

opensfm/reconstruction.py
```diff
@@ -64,6 +64,11 @@ def _add_camera_to_bundle(ba, camera, camera_prior, constant):
             camera.id, camera.focal, camera.k1, camera.k2,
             camera_prior.focal, camera_prior.k1, camera_prior.k2,
             constant)
+    elif camera.projection_type == 'dual':
+        ba.add_dual_camera(
+            camera.id, camera.focal, camera.k1, camera.k2,
+            camera.focal_prior, camera.k1_prior, camera.k2_prior,
```
paulinus: priors are now in `camera_prior`
```suggestion
            camera_prior.focal, camera_prior.k1, camera_prior.k2,
```

opensfm/exif.py
```diff
@@ -470,6 +470,23 @@ def camera_from_exif_metadata(metadata, data):
...
+        camera.id = metadata['camera']
+        camera.width = metadata['width']
+        camera.height = metadata['height']
+        camera.projection_type = pt
+        camera.focal = calib['focal']
+        camera.k1 = calib['k1']
+        camera.k2 = calib['k2']
+        camera.focal_prior = calib['focal']
+        camera.k1_prior = calib['k1']
+        camera.k2_prior = calib['k2']
```
paulinus: We can skip this.

opensfm/src/bundle/bundle_adjuster.h
```diff
@@ -128,6 +130,23 @@ struct BAFisheyeCamera : public BACamera{
   void SetK2(double v) { parameters[BA_CAMERA_K2] = v; }
 };
 
+struct BADualCamera : public BACamera{
+  double parameters[BA_CAMERA_NUM_PARAMS];
```
paulinus: Should we add a new enum with `BA_DUAL_CAMERA_NUM_PARAMS` so that `perspective` and `fisheye` cameras keep their 3 parameters instead of getting 4?

opensfm/mesh.py
```diff
@@ -20,12 +20,20 @@ def triangle_mesh(shot_id, r, graph, data):
...
+        return triangle_mesh_dual(shot_id, r, graph)
     elif shot.camera.projection_type in ['equirectangular', 'spherical']:
         return triangle_mesh_equirectangular(shot_id, r, graph)
     else:
         raise NotImplementedError
 
 
+def triangle_mesh_dual(shot_id, r, graph):
+    shot = r.shots[shot_id]
+    cam = shot.camera
```
paulinus: We do not seem to need this. Maybe we can skip that function and just call the fisheye version directly

PR #503, opened 2019-11-29
Refactor undistorted folder

paulinus:
This PR moves all files related to undistorted reconstructions to the `undistorted` subfolder. This includes undistorted images, masks, reconstructions, and depth maps. This makes it easier to have different undistorted versions of a reconstruction or to store them in other subfolders while decluttering the main folder.

The dataset methods related to undistorted data are moved to a new `UndistortedDataSet` class which keeps track of which subfolder the undistorted data is in.

This PR breaks the folder structure compatibility because depthmaps are stored in a different place. We should release a version before merging this so we can refer to it as the most recent using the older folder structure.

The new folder structure looks like tihs:
```
project/
├── config.yaml
├── images/
├── masks/
├── gcp_list.txt
├── exif/
├── camera_models.json
├── features/
├── matches/
├── tracks.csv
├── reconstruction.json
├── reconstruction.meshed.json
└── undistorted/
    ├── images/
    ├── masks/
    ├── tracks.csv
    ├── reconstruction.json
    └── depthmaps/
        └── merged.ply
```

The script `bin/migrate_undistort.sh` can be used to migrate existing datasets to the new folder structure.

### DIFF
--- a/bin/export_bundler
+++ b/bin/export_bundler
@@ -3,11 +3,8 @@ import os.path, sys
 sys.path.insert(0, os.path.abspath(os.path.join(os.path.dirname(__file__), "..")))
 import argparse
 
-import json
-import numpy as np
-
-import opensfm.dataset as dataset
-import opensfm.io as io
+from opensfm import dataset
+from opensfm import io
 
 
 if __name__ == "__main__":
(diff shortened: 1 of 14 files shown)


pierotofy:
:+1: very nice! This makes it easier to know which files are related to what.

PR #511, opened 2019-12-19
use IMREAD_UNCHANGED for undistort

pierotofy:
Hello :hand:

This PR proposes the addition of the `IMREAD_UNCHANGED` flag to the undistort command.

Reason:

When loading images with a single band or N-band (!= 3) camera, OpenCV will convert the image to a 3 channel image (losing the original band information).

Discussion for breaking changes and additional features:

This may break certain applications *if* such applications take as input single band or N-band images (N != 3) and make assumptions that the output of undistortion is always a 3 band RGB image. If that's the case, perhaps it might be beneficial to add a `--unchanged` or `--preserve-bands` flag to the undistort command. If that's the case, please let me know and I'll add it. :pray:

### DIFF
--- a/opensfm/commands/undistort.py
+++ b/opensfm/commands/undistort.py
@@ -97,7 +97,7 @@ def undistort_image_and_masks(arguments):
     logger.debug('Undistorting image {}'.format(shot.id))
 
     # Undistort image
-    image = data.load_image(shot.id, anydepth=True)
+    image = data.load_image(shot.id, unchanged=True, anydepth=True)
     if image is not None:
         max_size = data.config['undistorted_image_max_size']
         undistorted = undistort_image(shot, undistorted_shots, image,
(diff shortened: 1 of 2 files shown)


paulinus:
If we load the image `unchanged` but save it with the regular `io.imwrite`, will the order of the colors be preserved?

pierotofy:
The order is preserved.

```
from opensfm import io
from opensfm.dataset import DataSet

d = DataSet('/datasets/brighton2/opensfm')
i = d.load_image('/datasets/brighton2/images/DJI_0018.JPG', unchanged=True)
io.imwrite('/datasets/brighton2/out.jpg', i)
```

DJI_0018.JPG:

image

out.jpg:

image

If it were swapped it would look:

image

paulinus:
oh, I now see that `unchanged` is only for not dropping the alpha channel, which is clearly ok when undistorting. Thanks for the change.

PR #518, opened 2020-01-02
Add Python3 http.server command.

andrew-hardin:
Python3 has a slightly different command for launching an HTTP server. This pull request adds that command to `using.rst`.


paulinus:
thanks @andrew-hardin. i'll edit and let only the python3 version

PR #533, opened 2020-01-23
Set GPS DOP value to 15 (default) if EXIF reports it as 0

pierotofy:
I've found that certain cameras report a GPS DOP value of 0. I interpret this as meaning "no data" instead of indicating absolute precision (that doesn't seem possible).

I can post examples of problems with reconstructions when GPS DOP is set to zero if needed.

Let me know your thoughts?


GITNE:
While this PR is surely welcome and smart, there is a general problem with the `GPSDOP` tag because it is a relative value or factor of any device's absolute precision. And, no EXIF tag exists which could denote the absolute or maximum precision of a device. So, it is pretty much useless in the grand scheme of things, especially in the context of Mapillary where you have a myriad of devices with different precision capabilities.

YanNoun:
In any case, a `zero` value will cause issue in the bundle as we're gonna have zero-division.

Issue #554, opened 2020-02-27
Invalid NVM from undistort and export_visualsfm


linusmartensson:
The filename produced as output to an nvm file in https://github.com/mapillary/OpenSfM/blob/5825725706b3e86a488265e0d214924de21f0be3/opensfm/commands/export_visualsfm.py#L71 may currently contain spaces. 
This produces invalid NVM files that cannot be parsed by other editors, as the NVM file spec lacks escape characters for spaces, and separates its parameters by spaces.
This could be resolved by transforming the filenames in  https://github.com/mapillary/OpenSfM/blob/15d66bc284642b54ba7e2096923656efe12d823e/opensfm/dataset.py#L606
for example by hashing, filtering or b64-encoding the generated undistorted filename. 
Alternatively, this transformation could be done explicitly as part of the export step by copying the original files with issues to safe filenames, and referencing the copies in the export, preserving existing datasets without breaks. On systems that support symlinks, this could be done with efficient disk space usage.

YanNoun:
No answer in a while, and two PR merged, closing.

PR #556, opened 2020-02-28
Avoid broken NVM files by renaming input files

linusmartensson:
In https://github.com/mapillary/OpenSfM/issues/554 I noted that generated undistorted images with spaces in filenames break subsequent export steps and external tooling making use of NVM files. 
This transformation maintains all filenames without spaces, but ensures any filenames with spaces are corrected using a collision-avoiding naming strategy.

### DIFF
--- a/opensfm/dataset.py
+++ b/opensfm/dataset.py
@@ -604,7 +604,7 @@ class UndistortedDataSet(object):
         """Path of undistorted version of an image."""
         image_format = self.config['undistorted_image_format']
         if ' ' in image:
-            image.replace(' ', '') + hex(hash(image))
+            image = image.replace(' ', '') + hex(hash(image))
         filename = image + '.' + image_format
         return os.path.join(self._undistorted_image_path(), filename)
 
--- a/opensfm/dataset.py
+++ b/opensfm/dataset.py
@@ -603,6 +603,8 @@ class UndistortedDataSet(object):
     def _undistorted_image_file(self, image):
         """Path of undistorted version of an image."""
         image_format = self.config['undistorted_image_format']
+        if ' ' in image:
+            image.replace(' ', '') + hex(hash(image))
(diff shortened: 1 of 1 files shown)


YanNoun:
Hi,

Thank for your contribution. We understand that NVM file format doesn't support spaces in the filenames. However, there might be a better way to handle this issue, apart from changing OpenSfM implementation by adding a name mangling scheme.

The best would be not to have space in your filename. Or maybe you could add filename handling in NVM when its used by OpenDroneMap (assuming your use case is OpenSfM within OpenDroneMap). I'm pinging @pierotofy to know what's the policy regarding spaces-in-filenames/NVM.

Regards,

Yann

pierotofy:
Hey @YanNoun :hand: yes it could be handled on the ODM side before feeding the input to OpenSfM. It would be a bit more elaborate, as ODM tries not to fiddle with the files given as input from users (we'd have to make a copy to enable non-destructive behavior) and track invalid references from input GCPs, plus a few other edge-cases.

I'm OK with this not being included in OpenSfM, it's pretty specific to ODM and we can simply add it to our fork if needed. Although perhaps it might be useful to others. :+1:

linusmartensson:
In that case, could it be made an explicit error or warning in opensfm,
possibly with a documented explanation? The part I'm objecting to here is
primarily that it's a hidden error requiring understanding of the file
format specifics in nvm. Making it explicit would stop user errors
efficiently.

If you have a preference, I could put down the time on a PR. :)

Den tis 3 mars 2020 16:40Piero Toffanin <notifications@github.com> skrev:

paulinus:
@linusmartensson, as you suggested, I've added an error message that also stops the export when there are spaces in the image names. https://github.com/mapillary/OpenSfM/commit/fa022f369d660e73e45388914438292dd64af3e7

PR #557, opened 2020-03-03
Failure on unsorted data in apply_transformations

linusmartensson:
The original use of groupby expects a sorted transformations.keys() iterable. When this is not true, the function will repeatedly reload the original dataset and discard random alignments.
This results in alignment failures between submodels of large reconstructions.


YanNoun:
That's a tricky one, thanks a lot !

PR #565, opened 2020-03-23
Catch division by zero

pierotofy:
Hello :hand:

I've found a camera (DJI Zenmuse Z30) that reports "bad" `FocalPlaneXResolution` values of `0`, triggering a division by zero error.

Since `extract_sensor_width` can return `None` I think it should return that instead of failing.

Related: https://github.com/OpenDroneMap/ODM/issues/1089

### DIFF
--- a/opensfm/exif.py
+++ b/opensfm/exif.py
@@ -222,8 +222,10 @@ class EXIF:
         if not mm_per_unit:
             return None
         pixels_per_unit = get_tag_as_float(self.tags, 'EXIF FocalPlaneXResolution')
-        if pixels_per_unit == 0:
-            return None
+        if pixels_per_unit <= 0:
+            pixels_per_unit = get_tag_as_float(self.tags, 'EXIF FocalPlaneYResolution')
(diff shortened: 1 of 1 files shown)


opensfm/exif.py
```diff
@@ -222,6 +222,8 @@ def extract_sensor_width(self):
         if not mm_per_unit:
             return None
         pixels_per_unit = get_tag_as_float(self.tags, 'EXIF FocalPlaneXResolution')
+        if pixels_per_unit == 0:
+            return None
```
GITNE:
```suggestion
        if pixels_per_unit <= 0:
            pixels_per_unit = get_tag_as_float(self.tags, 'EXIF FocalPlaneYResolution')
            if pixels_per_unit <= 0:
                return None
```
If for some reason `FocalPlaneXResolution` is `0` or is broken we should try `FocalPlaneYResolution` before we finally return `None`. Of course, this assumes, like before, that pixels are square shaped (which on the vast majority of photo sensors they are).

PR #573, opened 2020-04-07
feat: C++ tracks manager

YanNoun:
This PR aims at replacing the original tracks graph based on `networkx` `Graph` by a custom C++ class in a newly created `sfm` modules. More specifically, we did the following :
 - Creation of `sfm` module
 - Re-structuring of the tests : each module has its own `test` folder and e should aim at gradually moving testing in C++.
 - Tracks manager and keypoint C++ classes
 - Renaming of all `XXX_tracks_graph` with `XXX_tracks_manager`

As a refactor-bonus, we also moved remaining exports to the command folder, so they can be used using `opensfm export_XXX`. They are now also exercised through testing, so unlikely to b broken in the future.

Note that we still use `networkx` `Graph` for the inliers graph, we this will also be replaced with a custom structure in the future.


opensfm/src/sfm/tracks_manager.h
```diff
@@ -0,0 +1,259 @@
...
+
+#include <fstream>
+#include <map>
+#include <unordered_map>
+#include <vector>
+
+class TracksManager {
+ public:
+  void AddTrack(const TrackId& id, const std::unordered_map<ShotId, Keypoint>& track) {
+    for (const auto observation : track) {
```
fabianschenk, line 14: const auto observation&

opensfm/src/sfm/tracks_manager.h
```diff
@@ -0,0 +1,259 @@
...
+    }
+  }
+
+  void AddObservation(const ShotId& shot_id, const TrackId& track_id, const Keypoint& observation) {
+    tracks_per_shot_[shot_id][track_id] = observation;
+    shot_per_tracks_[track_id][shot_id] = observation;
+  }
+
+  std::vector<ShotId> GetShotIds() const {
+    std::vector<ShotId> shots;
```
fabianschenk, line 26: shots.reserve(tracks_per_shot_.size())

opensfm/src/sfm/tracks_manager.h
```diff
@@ -0,0 +1,259 @@
...
+    for (const auto& it : tracks_per_shot_) {
+      shots.push_back(it.first);
+    }
+    return shots;
+  }
+
+  std::vector<TrackId> GetTrackIds() const {
+    std::vector<TrackId> tracks;
```
fabianschenk, line 34: Reserve to avoid repeated memory allocation: tracks.reserve(shot_per_tracks_.size())

paulinus:
@YanNoun, I've added a dockerfile for ubuntu 19.10 so that we can reproduce the double free issue I'm seeing locally.

Running the following reproduces the problem:
```
docker build -t opensfm.ubuntu1910 -f Dockerfile.ubuntu1910 .
docker run -ti opensfm.ubuntu1910 bin/opensfm_run_all data/berlin
```

paulinus:
For some mysterious reason, adding a missing `return true` statement https://github.com/mapillary/OpenSfM/pull/573/commits/3c9bdf97ead08ea12fed0f66c954397e3da1766d fixed the double free bug in ubuntu 19.10

PR #584, opened 2020-04-19
Extract EXIF metadata using all processors

hy9be:
Use multiprocessing

### DIFF
--- a/opensfm/commands/extract_metadata.py
+++ b/opensfm/commands/extract_metadata.py
@@ -1,6 +1,8 @@
 import copy
 import logging
 import time
+from multiprocessing import Pool, Manager, cpu_count
+from functools import partial
 
 from opensfm import dataset
 from opensfm import exif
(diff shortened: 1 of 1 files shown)


GITNE:
The speed up you are trying to achieve with your implementation is naive because it is based on naive assumptions and will probably even lead to a slowdown on the vast majority of systems. The problem is that this is not necessarily faster than a single thread code path because the expected performance when reading data from multiple **files** simultaneously depends heavily on the I/O characteristics of the storage controller on the target machine. When accessing multiple files simultaneously most operating systems, or more specifically the storage controller device driver, will try to do its best to share the a single resource, which is either the storage controller or the device driver itself, by slicing time slots.

Generally, the best approach is to fill up the command buffer of a storage controller (via the device driver) in a single-threaded one shot as fast as possible and then wait for completion callbacks on multiple CPUs to process the data.

If you really want to speed up meta data extraction as needed here, then you have to rewrite a lot more code than what you propose. You would need to load the TIFF headers of all files in a single thread in one go (fill up the command buffer asap) and then process the TIFF headers on multiple CPUs.

hy9be:
@GITNE I did some testing with a few datasets from ODM on 2 aws machines (c5.2xlarge, c5.4xlarge) provisioned with gp2 (SSD) hard disk. the multiprocessing could achieve almost linear improvement (with about 20% performance penalty). I think your point was right that the bottleneck in this step is I/O rather than CPU parallelism. But it looks the performance gain on SSD is still quite significant. 

AWS claims their gp2 SSD drive can reach 3000 IOPS (with 256KiB each for SSD), did that change the result?

GITNE:
Thank you for looking into this.
Generally speaking, even on SSDs Amdahl’s law still applies, so bare numbers of any specific device mean nothing here. Yet, it is true that you can possibly utilize more CPUs when loading initialization data from SSDs (as long as the data set per CPU is not larger than the throughput rate) than from hard disk drives. For one, because SSDs usually—due to their principle of construction—have lower access latency times and higher throughput rates than HDDs. Secondly, they often can also process more I/O ops per unit of time than HDDs. But, an upper limit still exists, which is dictated by the data set size per CPU and the throughput rate (plus access time but it is negligible for the sake of simplicity). So, I am pretty sure that the gains you might have observed are more due to the fact that the code is loading initialization data from an SSD rather then due to the algorithm scaling.

Your investigation and testing is appreciated. However, we have to be careful here not to compare apples to oranges, but to compare apples to apples. So, proper testing is paramount here.

In order to achieve reproducible results, please turn off the system file cache on the drive where the image files are located before you run the tests. Disable host caching on the VM instance. Make sure you are not sharing a CPU with other VMs, or better yet just run bare. Set the CPU (cores) to a fixed frequency. Then, try testing with 10, 100, and 1000 files, or even more, three to five times each while using a stable interrupt free timer, like the high resolution event timer, to measure timing results (note that although the machine time stamp counter is stable on a fixed frequency CPU, it is not interrupt free). Record the timings in a table. Happy testing! The results are surely going to be interesting. :wink:

PR #585, opened 2020-04-20
Export colmap features descriptors

santelelle:
The line for exporting colmap features descriptors was missing.
I also set the default add_two_view_geometry config=3 (fundamental matrix for colmap)


mlopezantequera:
@santelelle did you have a look at the scale and orientation of the features? that might be the cause of it not working

PR #586, opened 2020-04-21
Annotation GUI added

arajzd:
I have added the developed tool for annotating GCPs with the ability to render epipolar lines.


annotation_gui_gcp/Database.py
```diff
@@ -0,0 +1,178 @@
...
+
+    def get_path(self):
+        return self.path
+
+    def get_seqs(self):
+        return self.seqs
+
+    def init_points(self, points):
+        for point in points:
+            point_id, observations = point.values()
```
paulinus, line 65: This fails when the point dictionary has more than 2 entries. Better to access each field explicitly
```suggestion
            point_id, observations = point['id'], point['observations']
```

paulinus:
@arajzd, can the `vectors` folder can be removed since the icons are not used?

paulinus:
Looks good. Thanks so much for contributing this, @arajzd !

PR #594, opened 2020-05-06
feat: c++ camera models

YanNoun:
This PRs aims at replacing `OpenCV` usage for camera projection and distortion with our own implementation. To do so, we had to :
 - Add a `C++` implementation of current camera models supported by `OpenSfM`
 - Extend the functionality of the `NewtonRaphson` root finding to support the `n-D` case
 - Add corresponding unit tests in the `C++` part
 - Plug those camera models in `OpenSfM` `python` code

As a bonus :
 - We fix the dual camera implementation

PR #601, opened 2020-05-20
feat: camera models in bundle

YanNoun:
This PR aims at refactoring the bundle adjustment so that it uses the new camera class (see https://github.com/mapillary/OpenSfM/pull/594) for both API and internal cost function. To do so and maintain the same level of performance (more tricky than anticipated !), we had to :

(for the cameras)
 - Templated all functors in `camera_functions` so that they can be used by `Jet` 
 - Refactored all functors in `camera_functions` so they use raw types instead of Eigen types, mostly for performance issues (20-30%)
 - Refactor the `Camera` class by storing and exposing individual parameters along with their types. The class still stays ubiquitous but exposed in a way it's relevant for the considered camera model.
 - Add new cost functions `ReprojectionErrorXX` that directly use the `camera_functions` functions. Camera code is now at  a single place.
 

(for the bundle)
 - Add a new common structure and API `BAData` for getting any type as a contiguous vector
 - Refactor all cost instanciation accordingly to the refactored camera and cost functions. Most noticeably, we leverage the `Dispatch` function in `camera_functions` so that no more `switch/case` are needed.

As a bonus, we :
 - Refactored existing cost function by using `types.h` `typedefs`


opensfm/src/geometry/python/pybind.cc
```diff
@@ -22,69 +22,165 @@ PYBIND11_MODULE(pygeometry, m) {
...
+              camera = Camera::CreatePerspectiveCamera(values.at(Camera::Parameters::Focal),
+                                                        values.at(Camera::Parameters::K1),
+                                                        values.at(Camera::Parameters::K2));
+              break;
+            }
+            case ProjectionType::BROWN:{
+              Vec2d principal_point = Vec2d::Zero();
+              principal_point << values.at(Camera::Parameters::Cx), values.at(Camera::Parameters::Cy);
+              VecXd distortion(5);
+              principal_point << values.at(Camera::Parameters::K1), values.at(Camera::Parameters::K2),
```
fabianschenk, line 150: Should be **distortion** !!

opensfm/src/geometry/python/pybind.cc
```diff
@@ -22,69 +22,165 @@ PYBIND11_MODULE(pygeometry, m) {
...
+
+          std::vector<double> disto_values;
+          const auto disto_types = {Camera::Parameters::K1, Camera::Parameters::K2, Camera::Parameters::K3, Camera::Parameters::P1, Camera::Parameters::P2};
+          for (const auto type : disto_types) {
+            auto find_param = values_map.find(type);
+            if (find_param != values_map.end()) {
+              disto_values.push_back(find_param->second);
+            }
+          }
+          VecXd distorsion(disto_values.size());
```
paulinus, line 93: distortion

opensfm/src/geometry/camera_functions.h
```diff
@@ -1,338 +1,497 @@
...
  * pattern PROJ - > DISTO -> AFFINE. However, its is not mandatory for any
  * camera model to follow it. You can add any new camera models as long as it
  * implements the Forward and Backward functions. */
+
+/* Here's some trait that defines where to look for parameters that follows the
+ * generic scheme */
+template <class PROJ, class DISTO, class AFF>
+struct ParametersIndexTraits {};
+template <>
+struct ParametersIndexTraits<PerspectiveProjection, Disto24, UniformScale> {
```
paulinus, line 368: you can put that in a single trait as you have shown me

opensfm/src/geometry/camera_functions.h
```diff
@@ -1,338 +1,497 @@
...
+struct SizeTraits<DualCamera> {
+  static constexpr int Size = 4;
+};
+template <>
+struct SizeTraits<SphericalCamera> {
+  static constexpr int Size = 1;  // Dummy as we cant' allocate zero-size vector
+};
+
+/* Struct that can help overcomes the member template specialization */
+struct DispatchHelper {
```
paulinus, line 478: Let's add that only when needed

opensfm/src/bundle/src/bundle_adjuster.cc
```diff
@@ -484,6 +438,99 @@ struct BAParameterBarrier {
...
+  using Type = ReprojectionError2D;
+  static constexpr int Size = 2;
+};
+template <>
+struct ErrorTraits<SphericalCamera> {
+  using Type = ReprojectionError3D;
+  static constexpr int Size = 3;
+};
+
+struct DispatchCostFunctionPush : public DispatchHelper {
```
paulinus, line 452: Maybe call it `AddReprojectionErrorResidualBlock` or similar

PR #606, opened 2020-05-25
feat: C++ datastructures draft PR

fabianschenk:
This PR contains new C++ data structures for SfM and SLAM and is still WIP.
Short summary:
- Points, Shots, Cameras, and Observations are now written in C++ and can be accessed via Python
- `reconstruction` and `graph_inlier` are now stored in one data structure called `reconstruction`
- the old reconstruction` interface still works with small modifications

Discussion topics:
- API naming
- variable naming (e.g. points vs landmarks)
- `add_shot`, currently creates a shot based on another shot. Maybe `create_shot_from_shot` is better but very long
- can `main_shot` and `unit_shot` be discarded?

Last TODOs:
- [x] Add a `PanoShotView`
- [x] Fix the plotting functions` plot*.py`


GITNE:
> I think this commit may be the culprit behind broken bearing in reconstrunction.

Oh, this has not been merged yet. I take it back.

opensfm/commands/bundle.py
```diff
@@ -31,7 +31,8 @@ def run(self, args):
         gcp = data.load_ground_control_points()
 
         for reconstruction in reconstructions:
-            orec.bundle(graph, reconstruction, camera_priors, gcp, data.config)
+            # orec.bundle(graph, reconstruction, camera_priors, gcp, data.config)
```
YanNoun, line 34: Some left-overs to remove here.

opensfm/exif.py
```diff
@@ -422,6 +422,7 @@ def extract_exif(self):
 def hard_coded_calibration(exif):
     focal = exif['focal_ratio']
     fmm35 = int(round(focal * 36.0))
+    print("exif: ", exif)
```
YanNoun, line 425: Some left-overs to remove here.

opensfm/commands/export_visualsfm.py
```diff
@@ -7,7 +7,8 @@
 from opensfm import dataset
 from opensfm import transformations as tf
 from opensfm import io
-from opensfm import types
+# from opensfm import types
```
YanNoun, line 10: Idem

opensfm/io.py
```diff
@@ -57,27 +58,39 @@ def camera_from_json(key, obj):
...
-    metadata.gps_dop = obj.get("gps_dop")
-    metadata.gps_position = obj.get("gps_position")
-    metadata.skey = obj.get("skey")
-
-    shot = types.Shot()
-    shot.id = key
+    metadata = pymap.ShotMeasurements()
+    if obj.get("orientation") is not None:
+        metadata.orientation.value = obj.get("orientation")
+    if obj.get("capture_time") is not None:
```
YanNoun, line 73: I wonder if there would be a way to translate  some C++14 `Optional<T>` to corresponding `python` type or `None` :scratching_head_smiley:

opensfm/io.py
```diff
@@ -57,27 +58,39 @@ def camera_from_json(key, obj):
...
+        metadata.capture_time.value = obj.get("capture_time")
+    if obj.get("gps_dop") is not None:
+        metadata.gps_accuracy.value = obj.get("gps_dop")
+    if obj.get("gps_position") is not None:
+        metadata.gps_position.value = obj.get("gps_position")
+    if obj.get("skey") is not None:
+        metadata.sequence_key.value = obj.get("skey")
+    if obj.get("accelerometer") is not None:
+        metadata.accelerometer.value = obj.get("accelerometer")
```
YanNoun, line 82: Couldn't we have a `json_to_pymap_metadata` as symmetric of the one of `L235` ?

opensfm/io.py
```diff
@@ -512,11 +533,13 @@ def mkdir_p(path):
 
 def open_wt(path):
     """Open a file in text mode for writing utf-8."""
+    # print("Trying to open for write", path)
     return io.open(path, 'w', encoding='utf-8')
```
YanNoun, line 538: Some left-overs.

opensfm/src/map/src/shot.cc
```diff
@@ -0,0 +1,233 @@
...
+namespace map
+{
+
+Shot::Shot(const ShotId& shot_id, const Camera* const shot_camera,
+           const geometry::Pose& pose)
+    : id_(shot_id),
+      shot_camera_(shot_camera),
+      slam_data_(this),
+      pose_(pose),
+      merge_cc(0), scale(0) {}
```
YanNoun, line 14: Default scale to 1.0 ?

opensfm/src/map/test/map_test.cc
```diff
@@ -0,0 +1,147 @@
...
+  auto* my_cam = map.CreateCamera(camera);
+  map.CreateCamera(camera2);
+  for (auto i = 0; i < n_shots; ++i)
+  {
+    const auto shot_id = "shot"+std::to_string(i);
+    auto* shot = map.CreateShot(shot_id, i < n_shots/2 ? camera.id : camera2.id);
+    ASSERT_EQ(shot, map.GetShot(shot_id));
+    ASSERT_EQ(shot->id_, shot_id);
+    ASSERT_EQ(shot->unique_id_, i);
+    ASSERT_TRUE(map.HasShot(shot_id));
```
YanNoun, line 51: Would it be possible to have a fixture and split this test so we have separate tests for creation and deletion ?

opensfm/src/map/test/map_test.cc
```diff
@@ -0,0 +1,147 @@
...
+  for (auto i = 0; i < n_points; ++i)
+  {
+    Eigen::Vector3d pos = Eigen::Vector3d::Random();
+    const auto lm_id = std::to_string(i);
+    auto* lm = map.CreateLandmark(std::to_string(i), pos);
+    ASSERT_EQ(pos, lm->GetGlobalPos());
+    ASSERT_EQ(lm, map.GetLandmark(lm_id));
+    ASSERT_EQ(lm, map.CreateLandmark(std::to_string(i), pos));
+    ASSERT_EQ(lm->unique_id_, i);
+    ASSERT_EQ(map.NumberOfLandmarks(), i+1);
```
YanNoun, line 91: Same as above. Having smaller tests makes it easier in case of test failure.

opensfm/src/sfm/observation.h
```diff
@@ -15,4 +15,16 @@ struct Observation {
...
+
+  Observation(double x, double y, double s, uint8_t r, uint8_t g, uint8_t b, size_t id,
+            float _angle, float _resp, float _size, int _class_id)
+  : point(x, y), scale(s), color(r, g, b), id(id),
+    angle(_angle), response(_resp), size(_size), class_id(_class_id)
+    { }
+  
+  float angle;
+  float response;
+  // int octave; //this is probably the scale
```
YanNoun, line 27: You can remove it, we don't use it (and you're right, it is the inverse scale ;))

opensfm/test/test_types.py
```diff
@@ -268,12 +927,246 @@ def _get_dual_camera():
...
+    pose.set_rotation_matrix(R_wc)
+    assert np.allclose(R_wc, pose.get_R_world_to_cam())
+    r_wc = Rotation.from_dcm(R_wc).as_rotvec()
+    assert np.allclose(r_wc, pose.get_R_world_to_cam_min())
+
+    R_wc = special_ortho_group.rvs(3)
+    pose = pygeometry.Pose()
+    r_wc = Rotation.from_dcm(R_wc).as_rotvec()
+    pose.rotation = r_wc
+
```
YanNoun, line 1076: That would be great if we could split those tests.

opensfm/src/map/pybind_utils.h
```diff
@@ -0,0 +1,213 @@
...
+
+/// Makes an python iterator over the keys (`.first`) of a iterator over pairs from a
+/// first and past-the-end InputIterator.
+template <return_value_policy Policy = return_value_policy::reference_internal,
+          typename Iterator,
+          typename Sentinel,
+          typename KeyType = decltype((*std::declval<Iterator>()).second),
+          typename... Extra>
+iterator make_value_iterator(Iterator first, Sentinel last, Extra &&... extra) {
+    typedef detail::sfm_iterator_state<Iterator, Sentinel, detail::RefIterator, Policy> state;
```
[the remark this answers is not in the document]
fabianschenk: Python can only create a key iterator and an item iterator (key, value) and in this case, I wanted to have a value iterator and some specific iterators, (e.g.  for references).

opensfm/test/test_types.py
fabianschenk: I just saw that there are a few redundant asserts and tests. I'll try to split the tests and maybe write a `helper` to avoid the many redundant asserts.

opensfm/src/map/src/map.cc
```diff
@@ -0,0 +1,402 @@
...
+    lm->RemoveObservation(shot);
+  }
+}
+
+Shot* Map::GetShot(const ShotId& shot_id)
+{
+  const auto& it = shots_.find(shot_id);
+  return (it != shots_.end() ? &it->second : nullptr);
+}
+Shot* Map::GetPanoShot(const ShotId& shot_id)
```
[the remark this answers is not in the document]
fabianschenk: It throws now, similar to the `TracksManager`.

PR #610, opened 2020-06-12
feat: analytical jacobian of cameras

YanNoun:
This PR aims at adding analytical computation of camera projection functions, with the hope, it will speed-up bundle adjustment process. To do so, we had :
 - Generalize camera functions as `CameraFunctors`
 - Implement local jacobians of such camera functors.
 - Implement helper that automatically chain the above local jacobian, along the function composition.
 - Re-use the same to re-express the generic camera models.
 - Add corresponding unit test that uses `Eigen` `autodiff` to check for correctness.

As a bonus, we changed the parameter indexing in the camera class so it matches the ordering of the camera functions composition. This is still cumbersome and needs to be changed.

### DIFF
--- a/opensfm/src/geometry/camera_functions.h
+++ b/opensfm/src/geometry/camera_functions.h
@@ -537,6 +537,20 @@ static void ComposeForwardDerivatives(const T* in, const T* parameters, T* out,
       sub_jacobian, current_jacobian, jacobian);
 }
 
+template <class T, class FUNC>
+static void ComposeForward(const T* in, const T* parameters, T* out) {
+  FUNC::template Forward<T>(in, parameters, out);
+}
+
+template <class T, class FUNC1, class FUNC2, class... FUNCS>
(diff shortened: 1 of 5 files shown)


opensfm/src/geometry/camera_functions.h
```diff
@@ -36,14 +36,39 @@ struct FisheyeProjection {
...
+                                 T* jacobian) {
+    // dx, dy, dz
+    constexpr int Stride = COMP_PARAM*ParamSize + InSize;
+    jacobian[0] = T(1.0) / point[2];
+    jacobian[1] = T(0.0);
+    jacobian[2] = -point[0] / (point[2] * point[2]);
+    jacobian[Stride] = T(0.0);
+    jacobian[Stride + 1] = jacobian[0];
+    jacobian[Stride + 2] = -point[1] / (point[2] * point[2]);
+    T* dummy = NULL;
```
pmoulon, line 68: c++11

PR #612, opened 2020-06-19
Handle malformed XMP

pierotofy:
Hello :hand:

This PR proposes the addition of a second attempt in order to fix faulty XML in XMP tags from cameras that have malformed/invalid XML. It doesn't change the original behavior of the function. It loads `BeautifulSoup` lazily since I don't think too many cameras have this problem. The `lxml` library is a requirement for beautifulsoup.

I hope this can be useful to make OpenSfM more stable. 🙏  Let me know if changes are needed?


YanNoun:
I'm fine with the changes, but I have a few requests :
 - Let's imports `BeautifulSoup` at the top of the file since it is a requirement anyway. Makes the code less obfuscated.
 - I had trouble reading and understanding the combination of `range` loop + `try/except`. Could we remove the loop and make each step explicit, even if longuer and has nested `try/except` ? It would be more readable and indicates this is a tricky step part per-se.

YanNoun:
@pierotofy another thing would be to have at the image that caused the need for this fix and/or the corresponding malformed `string` so we have proper unit testing for that one, as it gets tricky.

pierotofy:
Thanks for the recommendations @YanNoun I'll work on the changes and also to provide you with a test case.

pierotofy:
Suggestions implemented. :+1: 

Test case:
```
from opensfm.exif import EXIF

e = EXIF(open("./DJI_0018.JPG"))
print(e.xmp)
```

Input image (with corrupted XMP): https://uav4geo.com/static/downloads/DJI_0018.JPG

Before the fix the print statement should fail. After the fix it should work.

pierotofy:
Mm, is a CLA required now? 

Not opposed to signing one, just not sure if I want to sign it via Facebook. :)

pierotofy:
Signed the CLA through Facebook, but this still fails. :confused:

pierotofy:
CLA signed, rebased the branch, fixed travis CI.

Let me know if further changes are needed :pray:

pierotofy:
Not sure if there's still interest in merging this one, I can close it if not. :pray:

PR #618, opened 2020-07-09
Add additional parameters to the fisheye camera model

bensaxby:
This pr adds the following parameters to the fisheye camera model:
* aspect ratio
* principal point cx and cy,
* k3 and k4 radial distortion parameters.

### DIFF
--- a/opensfm/commands/undistort.py
+++ b/opensfm/commands/undistort.py
@@ -211,7 +211,7 @@ def perspective_camera_from_perspective(distorted):
 
 
 def perspective_camera_from_brown(brown):
-    """Create a perspective camera froma a Brown camera."""
+    """Create a perspective camera from a Brown camera."""
     camera = pygeometry.Camera.create_perspective(
         brown.focal * (1 + brown.aspect_ratio) / 2.0, 0.0, 0.0)
     camera.id = brown.id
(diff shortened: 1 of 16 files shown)


YanNoun:
Hi @bensaxby ! Thanks for your PR, is great to see people adding more camera models. The PR looks quite complete, except for one big blocker : we need to keep the current behaviour of `Fisheye` cameras. So, a workaround will be to add a new camera type dubbed `FisheyeExtended`, so we would have both fisheye types.

You can make the change by yourself, or I can take care of it on your branch, let me know.

Yann

bensaxby:
Thanks for reviewing. I've made the change you suggested. The original `fisheye` camera model should work exactly as before, and a new `fisheye_extended` camera model has been created which uses the new distortion model and affine transformation.

I haven't touched the Colmap import/export since I'm not able to test that easily, so that won't support the new `fisheye_extended` model yet. All the tests pass for me except the map test, but that also fails on master and is fixed separately in #629. I don't think everything is covered by the tests though. Are there any parts you think need more coverage?

YanNoun:
Hi @bensaxby ,

Thanks for the refactor, it is really nice work ! Kudos for adding the model to the viewer ! Regarding the tests, as long as the bearing/project-loop and analytical derivatives works, it means we have a consistent model. Whether it fits or not a real camera is another topic ;-)

If @paulinus has no other objection, I think we can merge it.

Yann

paulinus:
Looks great! Thanks for going this deep to add the camera model, @bensaxby

Do i understand well that this is the same model as in the [opencv fisheye camera](https://docs.opencv.org/3.4/db/d58/group__calib3d__fisheye.html) ?

bensaxby:
Thanks! That is the intention and I think it is implemented correctly. I started this when the camera models were in Python and used the OpenCV fisheye camera model directly. This pr should be the equivalent change, but for the new C++ camera models.

YanNoun:
Sorry (as I made the `_extended` suggestion) for the inconvenience @bensaxby , but in the end, we decided to rename `_extended` to `_opencv`, so more people can get to use `OpenSfM` with the "Hey, that's the same OpenCV model :-) that I'm already using !"

PR #628, opened 2020-08-13
Feat: bundle and bundle_local in C++

fabianschenk:
This PR introduces C++ versions of `bundle` and `bundle_local` to allow for a faster problem setup in BA avoiding copies between C++ and Python with the new data structures (see #606) Further, it removes the Python versions of the `GroundControlPoint` and `GroundControlPointObservation` and replaces it by a C++ version:
Summary of all new C++ features, where most can be accessed via `pybind`:
- `bundle` and `bundle_local`
- `shot_neighborhood` and `direct_shot_neighborhood`
- `GroundControlPoint` and `GroundControlPointObservation`
- `triangulate_gcp`
- `add_gcp_to_bundle`

The only thing missing is the automatic detection of the alignment.


opensfm/reconstruction.py
```diff
@@ -384,6 +210,7 @@ def direct_shot_neighbors(reconstruction, shot_ids,
                 common_points[neighbor] += 1
 
     pairs = sorted(common_points.items(), key=lambda x: -x[1])
+    print(pairs)
```
YanNoun, line 213: Debug stuff

opensfm/reconstruction.py
```diff
@@ -801,7 +630,10 @@ def resect_reconstruction(reconstruction1, reconstruction2, tracks_manager1,
 def add_observation_to_reconstruction(tracks_manager, reconstruction,
                                       shot_id, track_id):
     observation = tracks_manager.get_observation(shot_id, track_id)
+    n_bef = reconstruction.shots[shot_id].compute_num_valid_pts()
```
YanNoun, line 633: Couldn't it hurt performance if we dot his `compute` thing at each observation we add ?

opensfm/src/map/python/pybind.cc
```diff
@@ -178,7 +182,8 @@ PYBIND11_MODULE(pymap, m) {
                     &map::Shot::SetShotMeasurements,
                     py::return_value_policy::reference_internal)
       // .def_readwrite("metadata", &map::Shot::shot_measurements_)
-      .def_property("pose", &map::Shot::GetPose, &map::Shot::SetPose)
+      // .def_property("pose", &map::Shot::GetPose, &map::Shot::SetPose)
+      .def_property("pose", (const geometry::Pose& (map::Shot::*) () const)  &map::Shot::GetPose, &map::Shot::SetPose)
```
YanNoun, line 186: What's the difference ?

opensfm/src/map/src/ba_helpers.cc
```diff
@@ -0,0 +1,528 @@
...
+    if (shot_ids.find(shot) == shot_ids.end()) {
+      candidate_shots.insert(shot);
+    }
+  }
+
+  std::unordered_map<map::Shot*, size_t> common_points;
+  for (auto* pt : points) {
+    for (const auto& neighbor_p : pt->GetObservations()) {
+      auto* shot = neighbor_p.first;
+      if (candidate_shots.find(shot) != candidate_shots.end()) {
```
YanNoun, line 69: Why the use of candidate shots ? Couldn't you use `shot_ids` directly ?

opensfm/src/map/src/ba_helpers.cc
```diff
@@ -0,0 +1,528 @@
...
+    ba.AddCamera(cam.id, cam, cam_prior, fix_cameras);
+  }
+
+  for (const auto& pt_pair : map.GetAllLandmarks()) {
+    const auto& pt = pt_pair.second;
+    ba.AddPoint(pt.id_, pt.GetGlobalPos(), false);
+  }
+
+  // Alignemnt method
+  // TODO: Alignment method, auto and vertical
```
YanNoun, line 364: Couldn't you pass the align method to the C++ method directly ? That way, you just move the L132-135 out of `bundle()` to get the align method, et then you use it straight into the C++ code. That would involve less rewrite.

YanNoun (approved):
Looks good overall, see my comment of the alignment method inside `bundle()`

YanNoun:
Also don't forget to remove all unused code in `reconstruction.py`

opensfm/reconstruction.py
fabianschenk: It's not really necessary and can hurt performance with the linear datastructures. In the SfM case, we use a map and it's just a call to size().

I will remove it

opensfm/src/map/python/pybind.cc
fabianschenk: The difference is that I have a const and non-const getter for the pose and this specifies the const getter

opensfm/src/map/src/ba_helpers.cc
fabianschenk: Yes, that would have worked. Now, I already ported the alignment method to C++.

PR #644, opened 2020-09-15
feat: use more features for panoramas (production behaviour)

YanNoun:
This PR does what the title says.

### DIFF
--- a/opensfm/actions/detect_features.py
+++ b/opensfm/actions/detect_features.py
@@ -7,6 +7,7 @@ from opensfm import bow
 from opensfm import features
 from opensfm import io
 from opensfm import log
+from opensfm import pygeometry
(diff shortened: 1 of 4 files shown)


opensfm/commands/detect_features.py
```diff
@@ -68,10 +78,14 @@ def detect(args):
     logger.info('Extracting {} features for image {}'.format(
         data.feature_type().upper(), image))
 
-    start = timer()
+    config = deepcopy(data.config)
```
paulinus, line 81: Can we avoid doing the copy? Maybe passing the `size` and `min_frames` as a parameter to extract features. Or passing an `is_panorama` boolean.

paulinus (approved):

Issue #647, opened 2020-09-22
GPS Altitude related issues


originlake:
1. In pr #473, it aims to use DJI XMP gps data if available, but it only checks if '@drone-dji:Latitude' exists in xmp data. In most of the cases I observed, the DJI XMP only has '@drone-dji:AbsoluteAltitude' but no @drone-dji:Latitude/@drone-dji:Longitude. So '@drone-dji:AbsoluteAltitude' will not be used anyway.
```
def extract_altitude(self):
    if self.has_dji_xmp():
        altitude = self.extract_dji_altitude()
    elif 'GPS GPSAltitude' in self.tags:
        altitude = eval_frac(self.tags['GPS GPSAltitude'].values[0])
    else:
        altitude = None
    return altitude

def has_dji_xmp(self):
    return (len(self.xmp) > 0) and ('@drone-dji:Latitude' in self.xmp[0])

def extract_dji_altitude(self):
    return float(self.xmp[0]['@drone-dji:AbsoluteAltitude'])
```
2. GPS AltitudeRef is not considered when extracting GPS Altitude
Opensfm takes the GPSAltitude value directly without considering altitude reference, it works when altitude reference is above sea level, but it's wrong when altitude reference is below sea level, "100 m below sea level"="-100 m above sea level", and it will cause issues like upside down for the reconstruction 3d model. And it's also inconsistent with  '@drone-dji:AbsoluteAltitude' value in 1, here an example shows the altitude metadata from an DJI image taken below sea level.
```
GPS Altitude Ref                : Below Sea Level
Absolute Altitude               : -62.88
Relative Altitude               : +59.50
GPS Altitude                    : 62.8 m Below Sea Level
```

originlake:
Created PR to solve the above issues #648

PR #648, opened 2020-09-22
fix: GPS Altitude extraction

originlake:
This PR fix below two issues
1. Use Absolute Altitude as GPS Altitude if it exists in DJI XMP data
2. Correct parsing GPS Altitude value if the GPSALtitudeRef is below sea level

Check issue #647 for details


YanNoun (approved):

YanNoun:
Hi @originlake !

Thank you for your work on the EXIF. It is almost perfect, except for a few function names. Would you mind renaming a bit things, like `has_dji_xmp` to `has_xmp` and also having function like `Has_dji_altitude` and `has_dji_latlon`.

Thank you,

Yann

originlake:
Hi @YanNoun,

I've modified EXIF as requests, renamed has_xmp, added has_dji_altitude and has_dji_latlon. Please review.

Thanks

PR #653, opened 2020-10-05
fix: do not init glog

paulinus:
We were internally initializing the glog library. This causes issues when other python libs also init it because it should only be initialized once.

This PR removes the initialization completely. If that causes issues, we can expose an init glog method to python and call it on each `__main__` on the lib.


DodgySpaniard (approved):
Looks good. Last time I checked there was no sane way to check if glog has been initialised, so this would be the lesser evil.

GITNE:
Thank you @paulinus for fixing this. :+1: Too bad glog does not initialize per shared object (or module in Windows' speak) like any well written library which requires initialization. :disappointed: Furthermore, glog also depends on libunwind, which is a pain in the @$$ to build and run on Windows.

PR #659, opened 2020-10-19
Use undistorted_image_max_size for all undistortions outputs

pierotofy:
Just noticed that the undistort command `undistorted_image_max_size` does not apply to all outputs.

Not sure if this warrants a renaming of the setting to `undistorted_output_max_size` or something similar (they are all images after all). Also the default in `config.py` for `undistorted_image_max_size` is currently `100000` which is not `1000000000` (`1e9`), but I don't think folks are processing images that large so this shouldn't break existing workflows.

:pray:


paulinus (approved):
It makes sense to use the same limit for all exported image types. Thanks!

PR #660, opened 2020-10-19
Fix transformation variable conflict

pierotofy:
Hello :hand:

In `export_geocoords.py` the variable `transformation` is declared twice (once to hold the transform matrix and once in the function args). I think it's a bug and causes a stack trace during normal invocation:

```
  File "/code/SuperBuild/src/opensfm/opensfm/actions/export_geocoords.py", line 40, in run_dataset
    if transformation:
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```

Renaming one of the two variables fixes the issue.


paulinus (approved):
Thanks for the fix @pierotofy !

PR #665, opened 2020-10-23
Add OpenCV 4.3.x SIFT support

RashidLadj:
Add tests in context.py to determine version 4.3.x

Add code in features.py to utilise the SIFT calls `cv2.xfeatures2d.SIFT_create()` when using OpenCV 4.3.x or send message error `no attribute 'xfeatures2d'` when using OpenCV 4.0.x, 4.1.x, 4.2.x.


paulinus:
Hi, @RashidLad. The OPENCV3 var is true for OPENCV3 or later. So the change does not modify the logic, but it makes it clearer. Thanks!

paulinus (approved):

PR #676, opened 2020-11-19
refactor: use threading

YanNoun:
This PR aims at making OpenSfM uses `joblib`'s `threading` instead of `multiprocessing`/`loky` : due to GIL release in OpenSfM C++ part and as well in OpenCV, critical code can run concurrently.

We narrow a bit the GIL release sections as some crash could be observed, and also release GIL when computing camera bearings map for undistortion.

We also fix the `command_runner` import broken at db32545f6d6359bb170382ba2d61a6be52247941

### DIFF
--- a/opensfm/src/geometry/python/pybind.cc
+++ b/opensfm/src/geometry/python/pybind.cc
@@ -11,6 +11,15 @@
 
 namespace py = pybind11;
 
+std::pair<MatXf, MatXf>
+RunComputeCameraMapping(const Camera& from,
+                        const Camera& to,
+                        int width, int height){
+  py::gil_scoped_release release;
+  return ComputeCameraMapping(from, to, width, height);
+}
+
+
 PYBIND11_MODULE(pygeometry, m) {
   py::enum_<ProjectionType>(m, "ProjectionType")
       .value("PERSPECTIVE", ProjectionType::PERSPECTIVE)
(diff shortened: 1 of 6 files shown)


YanNoun:
@pierotofy we've seen quite some improvement on feature extraction and matching.

pierotofy:
Woah! That's big.

opensfm/src/geometry/python/pybind.cc
```diff
@@ -251,7 +260,7 @@ PYBIND11_MODULE(pygeometry, m) {
            py::return_value_policy::copy)
       .def("__deepcopy__", [](const Camera& c, const py::dict& d) { return c; },
            py::return_value_policy::copy);
-  m.def("compute_camera_mapping", ComputeCameraMapping);
+  m.def("compute_camera_mapping", RunComputeCameraMapping);
```
paulinus, line 263: There seem to be a nice syntax we can use to do this without declaring the extra fuction
https://pybind11.readthedocs.io/en/stable/advanced/misc.html#global-interpreter-lock-gil
```suggestion
  m.def("compute_camera_mapping", RunComputeCameraMapping, py::call_guard<py::gil_scoped_release>());
```

paulinus (approved):
Great to have this working and giving speed ups straight away.
Is dense already working with threading?

opensfm/src/geometry/python/pybind.cc
YanNoun: Done.

YanNoun:
@pierotofy so it's almost good to go as it is. We observed random crashes during `compute_depthmaps`, that we think we fixed as we couldn't observe it anymore, but feel free to test it by yourself on a big dataset with many threads.

pierotofy:
Thanks @YanNoun ! I will run some tests soon. I will also look at the behavior of CTRL+C and signaling, I remember it was one of the reason we added `loky` at the time (the program wouldn't respond to SIGINT). :+1:

pierotofy:
Everything works good! :+1: No issues and definitely notice speed improvements!

PR #679, opened 2020-11-27
[fix] Add omp includes and libs to CMakeLists.txt

fabianschenk:
To compile on some system, it is required to explicitly add the `Open MP` includes and libraries in the `CMakeLists.txt`.
This solves compilation errors such as:
```
fatal error: 'omp.h' file not found 
#include <omp.h>
```

### DIFF
--- a/opensfm/src/sfm/CMakeLists.txt
+++ b/opensfm/src/sfm/CMakeLists.txt
@@ -11,7 +11,12 @@ target_link_libraries(sfm
   PRIVATE
     foundation
 )
-target_include_directories(sfm PUBLIC ${EIGEN_INCLUDE_DIRS} ${CMAKE_SOURCE_DIR})
+target_include_directories(sfm
+    PUBLIC
+        ${EIGEN_INCLUDE_DIRS}
+        ${CMAKE_SOURCE_DIR}
+    PRIVATE
+        ${OpenMP_CXX_INCLUDE_DIR})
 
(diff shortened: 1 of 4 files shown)


opensfm/src/sfm/CMakeLists.txt
```diff
@@ -11,7 +11,12 @@ target_link_libraries(sfm
   PRIVATE
     foundation
 )
-target_include_directories(sfm PUBLIC ${EIGEN_INCLUDE_DIRS} ${CMAKE_SOURCE_DIR})
+target_include_directories(sfm
+    PUBLIC
+        ${EIGEN_INCLUDE_DIRS}
+        ${CMAKE_SOURCE_DIR}
+    PRIVATE
+        ${OpenMP_CXX_INCLUDE_DIR})
```
paulinus, line 19: When should we use public or private? I see openmp added in the private section here and in the public one in the akaze file below

fabianschenk: That's a really good question. I have been playing around with that a bit and tried to reduce the includes to a minimum. To avoid having `PRIVATE` includes in most `CMakeLists.txt`,  `foundation` now adds the includes/libary as `PUBLIC` and `akaze` and `vlfeat` are still `PRIVATE`.

fabianschenk:
Closing because it's already part of another diff

PR #706, opened 2021-02-15
Match undistorted image names with shot ids

paulinus:
Summary:
Since we do not know on which format the input images are, we were always adding `.jpg` to the undistorted image file name before saving them. This caused a miss-match between the undistorted images' file names and the shot IDs in the undistorted reconstruction.

We fix that by renaming the shot IDs to match the image filename.

Additionally, we do not append the file format to the file name if the filename already ends with that extension. This breaks file-level backwards compatibility.

We also add the `undistorted_shot_ids.json` file to the dataset that maps original shot IDs to the list of corresponding undistorted shot IDs.  Remember that a single shot can have multiple undistorted shots.

Differential Revision: D26443052

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,11 @@
 ### Breaking
  - Main datastructures moved to C++ with Python bindings
  - Drop Python 2 support.  OpenSfM 0.5.x is the latest to support Python2.
+ - Undistorted image file names only append the image format if it does not match the distorted image source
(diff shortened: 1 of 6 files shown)


paulinus:
@pierotofy, heads up that this will change the name of the undistorted images. Whenever it was `image.jpg.jpg` it will now be `image.jpg`.
If the source image was a png, for example, the undistorted image will be `image.png.jpg`.

pierotofy:
Good to know! Happy we're removing the double extension :)

PR #713, opened 2021-03-01
Update OpenMVS exporter to be compatible with latest version

pierotofy:
This PR updates the `export_openmvs` command to generate files compatible with the latest OpenMVS.

Note that these changes don't allow older version of OpenMVS to process the files exported with this PR, so this is a breaking change.

### DIFF
--- a/opensfm/actions/export_openmvs.py
+++ b/opensfm/actions/export_openmvs.py
@@ -31,8 +31,8 @@ def export(reconstruction, tracks_manager, udata, export_only):
             w, h = camera.width, camera.height
             K = np.array(
                 [
-                    [camera.focal * max(w, h), 0, w / 2.0],
-                    [0, camera.focal * max(w, h), h / 2.0],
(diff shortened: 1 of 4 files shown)


opensfm/actions/export_openmvs.py
```diff
@@ -31,12 +31,12 @@ def export(reconstruction, tracks_manager, udata, export_only):
             w, h = camera.width, camera.height
             K = np.array(
                 [
-                    [camera.focal, 0, (w - 1.0) / 2 / max(w, h)],
-                    [0, camera.focal, (h - 1.0) / 2 / max(w, h)],
+                    [camera.focal * max(w, h), 0, w / 2.0],
```
cdcseacave, line 34: the -1 is missing from the change, I think it should remain

cdcseacave (changes requested):
I think the principal point computation is wrong

pierotofy:
Thanks for the correction @cdcseacave !

PR #719, opened 2021-03-09
<WIP> Feature: add flask cad viewer

chengchunhsu:
Adding following features:
- An web-based CAD model viewer
- Communications between flask server and UI
- Supporting of key-point annotation on CAD model


mlopezantequera:
Merged directly

PR #730, opened 2021-03-24
Update Pillow requirement

paulinus:
Summary: There was a vulnerability warning that required update

Differential Revision: D27293459

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,7 +5,7 @@ joblib==0.14.1
 matplotlib
 networkx==1.11
 numpy
-Pillow==7.1.0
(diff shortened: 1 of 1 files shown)


kirberich:
@paulinus I think this commit broke builds for the library:

```
The conflict is caused by:
    The user requested Pillow>=8.1.1
    fpdf2 2.1.0 depends on Pillow<=8 and >=4
```

Updating to fpdf 2.4.2 fixes it, but I'm not sure if that'll cause any other problems. I'll send a pr later, but let me know if you think that's not the right fix!

PR #735, opened 2021-04-21
Windows Support

pierotofy:
Hello :hand:

This PR gets OpenSfM to compile and run on Windows!

The main points that have been addressed:
 - [x] Rewrote the memory functions in `context.py` to work on all three platforms (Windows, Mac, Linux). This adds a new small dependency (`vmem`).
 - [x] Added a `bin/opensfm.bat` script for invoking the program.
 - [x] Some minor compilation issues related to Visual Studio are addressed. The definition of `M_PI` is a bit.. hackish, but the alternative is to include:

```
#define _USE_MATH_DEFINES
#include <cmath>
```

In a lot of places where M_PI is referenced. I'm not sure that's better (but I can change it if you want). I'm also unsure of what side effects the `_USE_MATH_DEFINES` macro might do. :bomb: 

 - [x] Updated the docs with build instructions for Windows. 
 - [x] Installs `opencv-python` as a pip dependency on Windows only; this is because vcpkg builds of OpenCV don't provide a version with OpenCV's Python bindings. This only affects Windows environments (it won't install on Mac or Linux).

I hope this can be useful to others.


paulinus:
Hi @pierotofy, thanks for making it work with windows!
We generally try to not add dependencies. It is difficult for us to make the internal builds work because vmem is not available internally. Is there an alternative to adding the dependency?  Can a windows version of `memory_available` be implemented directly in `context.py`?

pierotofy:
Hey @paulinus :hand: I could just copy the dependency in the opensfm tree? It's just a handful of files. See https://github.com/pierotofy/vmem/tree/main/vmem

paulinus:
The license is one issue because we don't want copy-left dependencies. But the best would be to entirely avoid adding a dependency.
If i understand correctly, the only functionality we miss is to know how much memory is available on windows. If this can be coded in a single function within context or similar, it would be ideal.

pierotofy:
Done! All platform related memory functions are now self-contained in `context.py`.

PR #750, opened 2021-05-11
Bug: fix memory deallocation error in load_reconstruction_shots

pierotofy:
Hello :hand:

I've found a memory deallocation issue in the `load_reconstruction_shots` method. Strangely this seems to only show up on Windows.

The key issue is that the `ShotView` object (`partial_reconstruction.shots`) is assigned by reference to the `reconstruction_shots`  dictionary. When the `reconstruction` object gets deallocated (when exiting the for loop scope), the ShotView object remains in existance, but the shots referenced by it do not, causing a crash.

The fix involves manually making copies of each shot.

To reproduce the error, on Windows, executing the `align_submodels` command will fail (reference #735 for running split-merge on Windows).

### DIFF
--- a/opensfm/large/tools.py
+++ b/opensfm/large/tools.py
@@ -235,7 +235,10 @@ def load_reconstruction_shots(meta_data):
         reconstruction = data.load_reconstruction()
         for index, partial_reconstruction in enumerate(reconstruction):
             key = PartialReconstruction(submodel_path, index)
-            reconstruction_shots[key] = partial_reconstruction.shots
+            reconstruction_shots[key] = {}
+
+            for shot_id in partial_reconstruction.shots:
+                reconstruction_shots[key][shot_id] = partial_reconstruction.shots[shot_id]
 
     return reconstruction_shots
 


ins7itia:
@pierotofy Hello,
Thank you for posting it! I found similar behavior with my own data, but it still caused error in Linux (Segmentation fault).
Furthermore, even if I explicitly deep copied shots, gps info was lost.

I solved the issue by using copy.deepcopy for shots and assigned metadata(for gps info) explicitly:

    reconstruction = data.load_reconstruction()
    for index, partial_reconstruction in enumerate(reconstruction):
        key = PartialReconstruction(submodel_path, index)
        reconstruction_shots[key] = partial_reconstruction.shots
        reconstruction_shots[key] = {}

        for shot_id in partial_reconstruction.shots:
            reconstruction_shots[key][shot_id] = copy.deepcopy(partial_reconstruction.shots[shot_id])
            reconstruction_shots[key][shot_id].metadata = partial_reconstruction.shots[shot_id].metadata

    return reconstruction_shots

I think it is not a concrete solution. Do you have any ideas?

pierotofy:
I'm a bit unsure of what would be the "right" way to fix this; the whole concept of a "ShotView" is that it's a "view", a reference to other objects. If one copies the view, but doesn't copy the references, well, you get problems.

In my opinion this could probably be refactored to use a generator pattern, so that you don't need to create a dictionary at all, you just return the generator which references the partial reconstructions' shots. But haven't delved into it, as this seemed to fix the issue I was experiencing (on Windows at least).

pierotofy:
This has been solved in the latest `master` (I actually had to revert this patch to get the latest changes in `master` to work).

PR #772, opened 2021-07-02
Update fpdf to be compatible with required Pillow version

kirberich:
The python requirements currently have invalid versions of fpdf and Pillow:

```
The conflict is caused by:
    The user requested Pillow>=8.1.1
    fpdf2 2.1.0 depends on Pillow<=8 and >=4
```

I've updated fpdf here and the library seems to build fine, but if there's a better fix I'm happy to change it!

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,7 @@
 cloudpickle==0.4.0
 exifread==2.1.2
 flask==1.1.2
-fpdf2==2.1.0
+fpdf2==2.4.2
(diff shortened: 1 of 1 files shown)


fabianschenk:
Main version is newer than the one in the diff:
https://github.com/mapillary/OpenSfM/blob/2aabdcb1d545ea67f04ef0dbc53365adab724b1d/requirements.txt#L4
Closing this PR.

PR #774, opened 2021-07-13
feat: compensate gps bias

YanNoun:
This Diff adds GPS bias compensation, which is useful when both GPS and GCP alignments are used : one want to align on GCP, while compensating for any bias the GPS receiver might have. More formally, the GPS error which was :

`error = GPS - image position`

Now becomes :

`error = bias(GPS) - image position`

With bias being a similarity transform. This is also exported in the report in the GPS/GCP section. Finally, this is controlled by the option `bundle_compensate_gps_bias`.


YanNoun:
Pushed from internal worflow.

PR #787, opened 2021-09-02
Apple M1 changes

pierotofy:
Hey all :hand: 

While building and running OpenSfM on an Apple M1 I encountered a strange bug, which I narrowed down to to `geometry/triangulation.h:71`. Debugging a reconstruction showed that for some reason the compiler decided to assume std::abs should be of type `int` instead of type `T`, causing values to be truncated to `0` at every computation. 

I've explicitly passed the `<T>` template to that call and for good measure I've added the parameter type to other calls to `std::abs`. I don't think these are strictly required, but to be safe..


pierotofy:
Closing in favor of https://github.com/mapillary/OpenSfM/pull/788

PR #788, opened 2021-09-03
Apple M1 changes

pierotofy:
Hey all :hand: 

While building and running OpenSfM on an Apple M1 I encountered a strange bug, which I narrowed down to to `geometry/triangulation.h:71`. Debugging a reconstruction showed that for some reason the compiler decided to assume std::abs should be of type `int` instead of type `T`, causing values to be truncated to `0` at every computation.

### DIFF
--- a/opensfm/src/geometry/triangulation.h
+++ b/opensfm/src/geometry/triangulation.h
@@ -68,7 +68,11 @@ std::pair<bool, Eigen::Matrix<T, 3, 1>> TriangulateTwoBearingsMidpointSolve(
 
   const T eps = T(1e-30);
   const T det = A.determinant();
+  #ifdef __aarch64__
+  if (std::abs<T>(det) < eps) {
+  #else
   if (abs(det) < eps) {
+  #endif
     return std::make_pair(false, Eigen::Matrix<T, 3, 1>());
   }
   const auto lambdas = A.inverse() * b;


opensfm/src/geometry/triangulation.h
```diff
@@ -68,7 +68,11 @@ std::pair<bool, Eigen::Matrix<T, 3, 1>> TriangulateTwoBearingsMidpointSolve(
 
   const T eps = T(1e-30);
   const T det = A.determinant();
+  #ifdef __aarch64__
+  if (std::abs<T>(det) < eps) {
```
fabianschenk, line 72: @pierotofy 
Thanks for your diff. Could you remove the `#ifdef` and simply use the templated version as default:

```
if (std::abs<T>(det) < eps)
```

pierotofy: Hey @fabianschenk :hand: thanks for the review, I actually started out by just having the templated version, but Visual Studio doesn't like it (read: it doesn't compile) on Windows.

I could replace the define to a `#ifndef __MSVC__ `?

fabianschenk: Hi @pierotofy ,
Thanks for the quick response and also for testing on Windows.
Then I'll land it like it is. 

Thanks again for your contribution 👍

pierotofy: Pleasure! OpenSfM is a great software.

PR #795, opened 2021-09-14
Windows Support (#735)

YanNoun:
Summary:
Hello :hand:

This PR gets OpenSfM to compile and run on Windows!

The main points that have been addressed:
 - [x] Rewrote the memory functions in `context.py` to work on all three platforms (Windows, Mac, Linux). This adds a new small dependency (`vmem`).
 - [x] Added a `bin/opensfm.bat` script for invoking the program.
 - [x] Some minor compilation issues related to Visual Studio are addressed. The definition of `M_PI` is a bit.. hackish, but the alternative is to include:

```
#define _USE_MATH_DEFINES
#include <cmath>
```

In a lot of places where M_PI is referenced. I'm not sure that's better (but I can change it if you want). I'm also unsure of what side effects the `_USE_MATH_DEFINES` macro might do. :bomb:

 - [x] Updated the docs with build instructions for Windows.
 - [x] Installs `opencv-python` as a pip dependency on Windows only; this is because vcpkg builds of OpenCV don't provide a version with OpenCV's Python bindings. This only affects Windows environments (it won't install on Mac or Linux).

I hope this can be useful to others.

Pull Request resolved: https://github.com/mapillary/OpenSfM/pull/735

Differential Revision: D29959025

Pulled By: YanNoun

### DIFF
--- a/bin/opensfm.bat
+++ b/bin/opensfm.bat
@@ -1,0 +1,8 @@
+@echo off
+
+setlocal
+set OSFMBASE=%~dp0
+
+python "%OSFMBASE%\opensfm_main.py" %*
+
+endlocal
--- a/doc/source/building.rst
+++ b/doc/source/building.rst
@@ -78,6 +78,25 @@ Also, in order for Cmake to recognize the libraries installed by Brew, make sure
     When running OpenSfM on top of OpenCV version 3.0 the `OpenCV Contrib`_ modules are required for extracting SIFT or SURF features.
 
 
+Installing dependencies on Windows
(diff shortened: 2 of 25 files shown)


YanNoun:
@pierotofy we're on the verge of merging it. What's the purpose of the `opencv-python==4.5.1.48 ; sys_platform == "win32"` in the `requirements` ?

pierotofy:
The OpenCV python bindings (`import cv2`) are not included in vcpkg, so they need to be installed separately.

If we don't want to explicitly include this in the requirements.txt, we can always just add an install step in the docs to tell people to install opencv-python separately.

PR #796, opened 2021-09-17
Fix errors when 0 feature detected

originlake:
Hello, 
This is the fix to handle the case when 0 feature is detected in the image. Currently, only the HAHOG feature detector can provide  correct ndarray shape for empty array while others will cause errors. This fix will assign `points` and `desc` with shape `(0,4)` and `(0, 128)` respectively if 0 feature is detected.

Attached an [example image](https://drive.google.com/file/d/16-JZlyOOwFcgwRJOth3SxZ4U0yxVXf1V/view?usp=sharing) to test with this case.


jonasdlindner:
This issue may be fixed here: https://github.com/mapillary/OpenSfM/pull/850/

originlake:
Close as resolved

PR #806, opened 2021-10-15
Fix Windows memory deallocation problem in load_reconstruction_shots

pierotofy:
For some reason on Windows the ShotView returned by partial_reconstruction.shots contains references to invalid shot objects. I documented this on https://github.com/mapillary/OpenSfM/pull/750 and thought it was resolved in the latest version, but this doesn't seem to be the case.﻿

This PR adds the logic to make explicit copies of the shot objects (on Windows only) to allow the `align_submodels` command to work properly.

### DIFF
--- a/opensfm/large/tools.py
+++ b/opensfm/large/tools.py
@@ -7,6 +7,8 @@ import cv2
 import networkx as nx
 import numpy as np
 import scipy.spatial as spatial
+import copy
+import sys
 from networkx.algorithms import bipartite
 from opensfm import align
 from opensfm import context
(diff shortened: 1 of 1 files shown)


YanNoun:
Hi @pierotofy,

This long-standing issue should have been solved by https://github.com/mapillary/OpenSfM/pull/822/files . Could you confirm it fixes the issue on your side.

Thank you !

Yann

pierotofy:
Thanks @YanNoun, I will test and get back to you on this :pray: (it might be a few weeks). I will close this PR in the meanwhile with the assumption that it does work, if it doesn't I will reopen.

PR #838, opened 2021-12-17
Extract OPK angles from XMP

pierotofy:
Hello :hand:

This PR adds support for extracting omega/phi/kappa angles from images' XMP tags (if they are available).

I've tested this with the `reconstruct --algorithm triangulation` command on a few datasets with a Sensefly and DJI camera and seems to yield the correct results.

image

Test dataset: https://github.com/pierotofy/drone_dataset_sheffield_cross/

Hope this can be useful to others! :pray:

### DIFF
--- a/opensfm/exif.py
+++ b/opensfm/exif.py
@@ -9,6 +9,7 @@ import xmltodict as x2d
 from opensfm import pygeometry
 from opensfm.dataset_base import DataSetBase
 from opensfm.sensors import sensor_data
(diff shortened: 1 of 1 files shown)


summitbri:
@pierotofy, is it possible to have this PR also work for the DJI P1 sensor (M300 drone)?

abmmusa:
Does this work only for drones? Can I use OPK for cars/vehicles? Do I need to follow another convention to make this work for vehicles?

PR #863, opened 2022-02-09
Fix include path for vlfeat in opensfm

chi3x10:
Summary: Fix the problem caused by previous diff.

Differential Revision: D34100698


YanNoun:
Done here : b3bb7638337c81436f1c1c2ca4701e6eb361e7ab

Issue #864, opened 2022-02-11
Images not in Images Directory leads to ZeroDivisionError


jonasdlindner:
I accidentally misspelled directory `data/berlin/images` to `data/berlin/image`.
Instead of getting an error like:
```
Traceback (most recent call last):
   File "/source/OpenSfM/bin/opensfm_main.py", line 25, in <module>
     commands.command_runner(
   File "/source/OpenSfM/opensfm/commands/command_runner.py", line 37, in command_runner
     command.run(data, args)
   File "/source/OpenSfM/opensfm/commands/command.py", line 12, in run
     self.run_impl(data, args)
   File "/source/OpenSfM/opensfm/commands/detect_features.py", line 11, in run_impl
     detect_features.run_dataset(dataset)
   File "/source/OpenSfM/opensfm/actions/detect_features.py", line 28, in run_dataset
     max_queue_size, int(expected_mb / average_image_size(data))
 ZeroDivisionError: float division by zero
```
i would appreciate an more descriptive message like:
```
No Images found in data/berlin/images
```

PR #865, opened 2022-02-11
Raise Error when Image Path is empty

jonasdlindner:
Fixes https://github.com/mapillary/OpenSfM/issues/864


fabianschenk:
Hi @jonasdlindner ,

Thanks for submitting this PR. 
It breaks many of the tests (see [here](https://github.com/mapillary/OpenSfM/runs/5157353078?check_suite_focus=true)), so please try to refine it so that all the tests are green 😄 

Thanks again

Issue #802, opened 2021-10-09
Build issue: search paths for Eigen are hard coded in cmake file


rudyryk:
https://github.com/mapillary/OpenSfM/blob/343d1c349a008507f3b7e4be84e31e74c52f92f5/opensfm/src/cmake/FindEigen.cmake#L88

This prevents from having Eigen to be installed to custom location even if paths are explicitly provided via `CFLAGS` or `CXXFLAGS` etc.

In my case, I was using Nix packager and shell:

```bash
$ nix-shell -p python38Full python38Packages.opencv4 \
    python38Packages.setuptools python38Packages.sphinx \
    python38Packages.wheel cmake clang opencv suitesparse \
    blas lapack eigen ceres-solver

[nix-shell:~/OpenSfM]$ python setup.py build
```

Generated output:

```
Configuring for python 3.8...
-- Could NOT find OpenMP_C (missing: OpenMP_C_FLAGS OpenMP_C_LIB_NAMES)
-- Could NOT find OpenMP_CXX (missing: OpenMP_CXX_FLAGS OpenMP_CXX_LIB_NAMES)
-- Could NOT find OpenMP (missing: OpenMP_C_FOUND OpenMP_CXX_FOUND)
-- Found AMD library: /nix/store/hsw3xs8lxiklblc6lvdqrp3rpz3mh440-suitesparse-5.9.0/lib/libamd.dylib
-- Found AMD header in: /nix/store/43hac8g89rxcy8a8wjcq36ilg8bschl7-suitesparse-5.9.0-dev/include
-- Found CAMD library: /nix/store/hsw3xs8lxiklblc6lvdqrp3rpz3mh440-suitesparse-5.9.0/lib/libcamd.dylib
-- Found CAMD header in: /nix/store/43hac8g89rxcy8a8wjcq36ilg8bschl7-suitesparse-5.9.0-dev/include
-- Found COLAMD library: /nix/store/hsw3xs8lxiklblc6lvdqrp3rpz3mh440-suitesparse-5.9.0/lib/libcolamd.dylib
-- Found COLAMD header in: /nix/store/43hac8g89rxcy8a8wjcq36ilg8bschl7-suitesparse-5.9.0-dev/include
-- Found CCOLAMD library: /nix/store/hsw3xs8lxiklblc6lvdqrp3rpz3mh440-suitesparse-5.9.0/lib/libccolamd.dylib
[...]

Issue #871, opened 2022-02-19
Utilities diverge from the modern pybind11


SomeoneSerge:
The pinned pybind11 submodule is relatively old. In the upstream pybind11 there already are provided some of the utilities from `pybind_utils.h`, like `make_value_iterator`, `iterator_state`, but they have different signatures (e.g. 5 template parameters against 4). I hoped to build OpenSfM without submodules with the (newer) pybind11 version pinned outside the repo, but this would require modifications to the template definitions

- Would you consider updating to the newer pybind11 before the next release?
- Do you need the in-tree pybind11 in the first place?

Issue #870, opened 2022-02-19
Do you need the in-tree copy of gtest?


SomeoneSerge:
The distributed `FindGlog.cmake` and `FindGtest.cmake` diverge from the upstream's interface: the modern gtest and glog seem to link through the targets `glog::glog` and `GTest::gtest`, whereas you rely on the `GLOG_LIBRARY` variable. This is somewhat of a hindrance when trying to build OpenSfM with system-distributed versions of gtest/glog and creates a necessity for patching the CMakeLists before the build: https://github.com/NixOS/nixpkgs/pull/152957/files#diff-393b1ae08ea4ab1b6c409e106cdf13ef1ac32d9777af6744bb895bfaf37b9346R52

Do you need the custom copy of `gtest` and custom `Find*` files in the first place?

PR #872, opened 2022-02-19
CMake: target_link_libraries() the native exported targets of Eigen and Ceres

SomeoneSerge:
- [x] Upstream documentation suggests eigen be used through the `Eigen3::Eigen` target: https://eigen.tuxfamily.org/dox/TopicCMakeGuide.html
- [x] Eliminates the need for custom FindEigen.cmake and hard-coded paths
- [x] Makes it easier to package opensfm for distributions
- [x] Closes: https://github.com/mapillary/OpenSfM/issues/802
- [x] Checked `docker build -f Dockerfile.ceres2 .` and `nix build`
- [ ] `target_link_libraries` vs `target_include_directories` propagation rules: cf. bottom of the section https://cmake.org/cmake/help/latest/manual/cmake-buildsystem.7.html#transitive-usage-requirements

I needed these changes in the process of packaging OpenSfM for nixpkgs, which significantly simplifies the building and deployment compared to plain docker: https://github.com/NixOS/nixpkgs/pull/152957
You can see the build status, including the tests (currently failing to build for macos) at: https://github.com/SomeoneSerge/pkgs/actions

I'd hope to follow up with a few more changes that I found I needed, but I'd need your advice on how to go about them:
- Discovering system-distributed gtest?  https://github.com/mapillary/OpenSfM/issues/870
- Discovering system-distributed  pybind?  https://github.com/mapillary/OpenSfM/issues/871

Thank you!


opensfm/src/CMakeLists.txt
```diff
@@ -54,11 +54,21 @@ endif()
...
+find_package(Ceres)
 find_package(Gflags REQUIRED)
 find_package(Glog REQUIRED)
 
+# Ceres2 exposes Ceres::ceres target.
+# Ceres1 exposes just ceres.
+# - if there's no such target, cmake will convert it into -lceres
+# and the linker will fail
+if(Ceres_ceres_FOUND)
+    set(CERES_LIBRARIES Ceres::ceres)
```
SomeoneSerge, line 67: I'm still using (but also explicitly defining the `CERES_LIBRARIES` variable because one needs to `target_link_libraries` different targets, conditionally on the version of ceres

opensfm/src/bundle/CMakeLists.txt
```diff
@@ -22,25 +22,27 @@ target_link_libraries(bundle
...
 )
+
 if (LAPACK_FOUND)
     target_include_directories(bundle PRIVATE ${LAPACK_INCLUDE_DIRS})
 endif()
 if (SUITESPARSE_FOUND)
     target_include_directories(bundle PRIVATE ${SUITESPARSE_INCLUDE_DIRS})
 endif()
-target_include_directories(bundle PRIVATE ${CERES_INCLUDE_DIR} ${CMAKE_SOURCE_DIR})
+target_include_directories(bundle PRIVATE ${CMAKE_SOURCE_DIR})
```
SomeoneSerge, line 32: But otherwise preferring to reference only target names (`Ceres::ceres` in `target_link_libraries`), instead of lists of paths (`CERES_INCLUDE_DIR`)

opensfm/src/bundle/CMakeLists.txt
```diff
@@ -22,25 +22,27 @@ target_link_libraries(bundle
...
         test/bundle_data_test.cc
     )
     add_executable(bundle_test ${BUNDLE_TEST_FILES})
-    target_include_directories(bundle_test PRIVATE ${CMAKE_SOURCE_DIR} ${EIGEN_INCLUDE_DIRS} ${GMOCK_INCLUDE_DIRS})
+    target_include_directories(bundle_test PRIVATE ${CMAKE_SOURCE_DIR} ${GMOCK_INCLUDE_DIRS})
     target_link_libraries(bundle_test
                         PUBLIC
                         bundle
                         geometry
+                        Eigen3::Eigen
```
SomeoneSerge, line 45: It's `target_link_libraries` but it's exported like:

```
add_library(Eigen3::Eigen INTERFACE IMPORTED)
```

opensfm/src/CMakeLists.txt
SomeoneSerge: Hello! I'm not sure if it's possible: if we keep `FindCeres.cmake`, then `find_package(ceres ...)` a few lines above would probably always encounter the in-tree `FindCeres.cmake` before it checks the other directories in `CMAKE_PREFIX_PATH`. I'll try to research that

opensfm/src/bundle/CMakeLists.txt
SomeoneSerge: Hello! It would because we keep `target_link_libraries(... ${CERES_LIBRARIES})`, where `${CERES_LIBRARIES}` refer to an exported Ceres cmake target (it propagates the include directories). The ceres/Ceres::ceres target would be discovered by `find_package` through the `CMAKE_PREFIX_PATH` environment variable

SomeoneSerge: For example, Nix puts ceres in a "custom" location at `/nix/store/${hash}-ceres-solver/`, but cmake is able to discover it, because with [ceres in "inputs"](https://github.com/NixOS/nixpkgs/blob/c02944e17d0d8ea91c9dc8878dc93697f7098838/pkgs/development/python-modules/opensfm/default.nix#L78) Nix adds this path to `CMAKE_PREFIX_PATH` list

PR #972, opened 2023-01-03
Update numpy to 1.19.0

yanxke:
The default numpy version is too low and running the example in the documentation fails on Matplotlib because it requires 1.19.0 or higher.


fabianschenk:
Numpy and scipy were updated in https://github.com/mapillary/OpenSfM/pull/976
Closing this PR now

PR #1053, opened 2024-05-24
Remove dublicated SIFT-Dectector instance in feature matching

kielnino:
It seems to me that the SIFT detector is instantiated, but then not used, because the object is created again in the following while loop. This merge request removes the unnecessary code.
I have also made the missing parameters of the OpenCV-Sift implementation configurable.

### DIFF
--- a/opensfm/config.py
+++ b/opensfm/config.py
@@ -42,6 +42,12 @@ class OpenSfMConfig:
     sift_peak_threshold: float = 0.1
     # See OpenCV doc
     sift_edge_threshold: int = 10
+    # See OpenCV doc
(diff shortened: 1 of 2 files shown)


kielnino:
Maybe we should move the line `descriptor = detector` to run only once after the while-Loop, conditioned on the `context.OPENCVXX` flags?

kielnino:
@fabianschenk Do I need to adjust something to make the FB-internal tests pass?

And in addition do you have any thoughts about this?
> Maybe we should move the line `descriptor = detector` to run only once after the while-Loop, conditioned on the `context.OPENCVXX` flags?

PR #1090, opened 2025-02-27
Cones to merge

aliaksandr960:


aliaksandr960:
It is my first merge request to open-source repo. If it violate something, please provide a way, how it should be improved. I really not aware how to do it in a proper way. Sorry.

opensfm/pairs_selection_by_cones.py
```diff
@@ -0,0 +1,196 @@
+import numpy as np
+
+import numpy as np
```
frolui, line 3: Why do we need such duplication of imports in lines 1-20 @aliaksandr960 ?

PR #1096, opened 2025-06-19
Warn user if ceres loss function is not available

kielnino:
I set TukeyLoss without noticing, that it wasn't used. So in this PR `CreateLossFunction` throws a runtime_error if the loss-function is not implemented, instead of silently using nullptr.


kielnino:
Hi @fabianschenk, thank you for taking on the matter. I am not yet completely satisfied with the wording of the error message, as the lossfunction may be present in ceres, but is simply not handled in the if-else switch. This could confuse the user somewhat.

Issue #599, opened 2020-05-18
Fix: Arbitrary focal plane and length conversion to 35 mm film focal length


GITNE:
https://github.com/mapillary/OpenSfM/blob/2005aae849ae5c43e781c2ae8ced1d550e36675e/opensfm/exif.py#L63-L64

Apparently, OpenSfM assumes or works with 35 mm film focal length internally (I was not aware of that), which is not great but okay. Anyhow, the current conversion from an arbitrary focal plane and length to full-frame/35 mm film focal length is incomplete, to say the least. So, you get things [like this](https://www.mapillary.com/app?pKey=l5rhkrPiLb_I-09oL39yhg&focus=photo "traverse to the end").

To fix this we have to compute the true equivalent focal length in 35 mm film:

```python
focalLengthIn35mmFilm = focalLength /
    sqrt(focalPlaneWidth^2 + focalPlaneHeight^2) *
    sqrt(36.0^2 + 24.0^2)
```

We have to convert to 35 mm film via the focal plane’s hypotenuse because we have to take into account all the different aspect ratios which are out there.

Besides, I would also prefer evaluating `FocalLength`, `FocalPlaneXResolution`, `FocalPlaneYResolution` first before evaluating `FocalLengthIn35mmFilm` because `FocalLengthIn35mmFilm` is just a *lazy* imprecise equivalent of the true focal length; it is an integer value of type `Short` in the Exif metadata header. Exactly for that reason the fully precise `FocalLength`, `FocalPlaneXResolution`, `FocalPlaneYResolution` (and `FocalPlaneResolutionUnit`) Exif tags have been introduced after `FocalLengthIn35mmFilm` has been found or deemed unfit for many purposes. The amount of deviation caused by `FocalLengthIn35mmFilm`’s imprecision does not seem like much on a per image basis but it accumulates rapidly over many images (like OpenSfM is used). Note, that it is still okay for OpenSfM to use the 35 mm film focal length internally because we can compute fractions of millimeters in the metadata extraction step.

@paulinus Do you want me to create a PR or would you like to add a quick fix?

PR #1106, opened 2025-08-21
Fix 35mm conversion

YanNoun:
This PR fixes the 35mm focal <-> ratio focal conversion : 
 - 3/2 image ratio uses 36mm filback width
 - 4/3 uses a 34mm one

Branch is on top of optim-large

### DIFF
--- a/opensfm/exif.py
+++ b/opensfm/exif.py
@@ -3,6 +3,7 @@
 import datetime
 import logging
 from codecs import decode, encode
+import math
 from typing import Any, BinaryIO, Callable, Dict, List, Optional, Tuple, Union
 
 import exifread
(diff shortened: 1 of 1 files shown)


opensfm/exif.py
```diff
@@ -60,22 +61,52 @@ def get_tag_as_float(tags: Dict[str, Any], key: str, index: int = 0) -> Optional
...
+    focal35_or_ratio: float, width: int, height: int, inverse=False
+) -> float:
+    """
+    Convert focal length in 35mm film equivalent to focal ratio (and vice versa).
+    We follow https://en.wikipedia.org/wiki/35_mm_equivalent_focal_length
+    """
+    image_ratio = float(max(width, height)) / min(width, height)
+    is_32 = math.fabs(image_ratio - 3.0 / 2.0)
+    is_43 = math.fabs(image_ratio - 4.0 / 3.0)
+    if is_32 < is_43:
```
paulinus, line 74: The wikipedia article mentions a definition that works for any aspect ratio in the "Calculation" section

> "Converted focal length into 35 mm camera" = (Diagonal distance of image area in the 35 mm camera (43.27 mm) / Diagonal distance of image area on the image sensor of the [DSC](https://en.wikipedia.org/wiki/Digital_Camera)) × focal length of the lens of the DSC.

Should we use that instead so that we support any aspect ratio rather than picking between these two?
I guess the question is which definition do camera manufacturers use to fill the EXIF tag.

paulinus (changes requested):
The PR seem to have mixed the actual change with other unrelated commits.

opensfm/exif.py
JakeSmarter: You have made a valid observation. Thank you for bringing this up. However, imho the focal length conversion should be based on diagonal sensor size in order to correctly and transparently support all aspect ratios and sensor dimensions. I would suggest for your PR to focus on propagating this aspect though the rest of the code rather than trying to revert it or add separate code paths for each aspect ratio.
> I guess the question is which definition do camera manufacturers use to fill the EXIF tag.

Although it is not explicitly specified in the Exif spec, afaik they assume diagonal equivalency, exactly due to different sensor aspect ratios. I have tried to address this in #599.

JakeSmarter: > CIPA DCG-001-Translation- 2005-12-
> ## (2 – c) “Conversion of focal length of DSC lens into 35mm camera equivalent”
> The focal length of a DSC lens converted into that of 35mm camera lens which has the same input field angle as the DSC lens.
> Rules for notation
> * a) “Converted into 35mm camera” or similar wording should be noted.
> * b) The value is to be calculated using the following equation;
>   “Converted focal length into 35mm camera” = **Diagonal** distance of image area in the 35mm camera (43.27mm) ÷ **Diagonal** distance of image area on the image sensor of the DSC × focal length of the lens of the DSC
> * c) All of the numerical values can be rounded off to two decimal places.

JakeSmarter: I also do not see much reason for the `is_32` and `is_43` variables, unless something eludes me. If you are trying to determine the image orientation then test for the `Orientation` or `PixelXDimension`/`PixelYDimension` Exif tags. If these are not present then it is safe to assume that width ≥ height by standard definition. Again, unless an image has been deliberately cropped into an unusual aspect ratio than the original image or the sensor has non-square pixels. But, these are obscure cases.

PR #1108, opened 2025-08-21
Improve final point cloud with filtering

YanNoun:
This PR adds a point cloud cleaning step that removes points with high covariance ellipsoid, and isolated points. This greatly improves the final point cloud, especially if subsequent steps rely on it directly.

On a Mapillary dataset : CLEAN on LEFT, ORIGINAL on RIGHT
Screenshot From 2025-08-21 16-48-08
Screenshot From 2025-08-21 16-47-30

On an aerial dataset  : 
CLEAN
Screenshot From 2025-08-21 12-43-55
ORIGINAL
Screenshot From 2025-08-21 12-34-06

This branch is on top fix-gcp-handling

### DIFF
--- a/opensfm/src/sfm/src/tracks_helpers.cc
+++ b/opensfm/src/sfm/src/tracks_helpers.cc
@@ -8,8 +8,13 @@
 #include <vector>
 #include <cmath>
 #include <numeric>
+#include <iostream>
 #include <unordered_set>
 #include <algorithm>
+
+extern "C" {
+#include <third_party/vlfeat/vl/kdtree.h>
+}
 
 namespace sfm::tracks_helpers {
 std::unordered_map<map::ShotId, int> CountTracksPerShot(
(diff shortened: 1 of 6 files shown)


paulinus:
Unrelated to the PR: why is the focal length of the camera different in the left/right screenshots of the mapillary sequences?
Visually, from the camera pyramids, left seems to have a longer focal length than right.

opensfm/src/sfm/src/tracks_helpers.cc
```diff
@@ -60,4 +68,154 @@ void RemoveConnections(map::TracksManager& manager, const map::ShotId& shot_id,
...
+      if (!obs_ptr) {
+        continue;
+      }
+
+      cameras.push_back(*cam_ptr);
+      poses.push_back(*pose_ptr);
+      obs_vec.push_back(obs_ptr->point);
+    }
+
+    // First simple check based on angle between all pairs of raysmap
```
paulinus, line 128: We do this check before triangulating the points. Is the idea that after bundle that could have changed?

Issue #1083, opened 2024-12-11
BundleShotPoses possibly broken


kielnino:
I have a simple setup with some images to reconstruct, all taken by the same camera. As I understand the code, individual shots are added to a fake rig instance that only contains the one shot.

In the reconstruction step `grow_reconstruction` the newly added shots are passed to [bundle_shot_poses](https://github.com/mapillary/OpenSfM/blob/b72eb60ea132f02f98bd2b6d025234950cddb5c0/opensfm/reconstruction.py#L1448).

As the newly added shot has the *fake* rig_instance_id with the same id as the shot_id, `rig_instances_ids` now only contains the newly added shot. In [Line 481](https://github.com/mapillary/OpenSfM/blob/b72eb60ea132f02f98bd2b6d025234950cddb5c0/opensfm/src/sfm/src/ba_helpers.cc#L481) the  `rig_instances_ids` are iterated and the corresponding shots are added to the bundle problem (Line [496](https://github.com/mapillary/OpenSfM/blob/b72eb60ea132f02f98bd2b6d025234950cddb5c0/opensfm/src/sfm/src/ba_helpers.cc#L496)).

[As part of this](https://github.com/mapillary/OpenSfM/blob/b72eb60ea132f02f98bd2b6d025234950cddb5c0/opensfm/src/sfm/src/ba_helpers.cc#L501) it is checked if the found shot_id is in the list of the newly-added shots (which are supposed to be bundled), and this is always the case. So all shots are fixed and the BA only consists of constant blocks. See the ceres summary full-report:

```
Solver Summary (v 2.1.0-eigen-(3.4.0)-lapack-eigensparse-no_openmp)

                                     Original                  Reduced
Parameter blocks                           67                        0
Parameters                                216                        0
Residual blocks                            65                        0
Residuals                                 140                        0

Minimizer                        TRUST_REGION

Sparse linear algebra library    EIGEN_SPARSE
Trust region strategy     LEVENBERG_MARQUARDT

                                        Given                     Used
Linear solver                        DENSE_QR   SPARSE_NORMAL_CHOLESKY
Threads                                    20                       20
Linear solver ordering              AUTOMATIC                AUTOMATIC

Cost:
Initial                          9.919183e-01
Final                            9.919183e-01
Change                           0.000000e+00

Minimizer iterations                       -2
Successful steps                           -1
Unsuccessful steps                         -1

Time (in seconds):
Preprocessor                         0.000770

[...]

PR #1115, opened 2026-02-06
Correct definition of is_fixed in BundleShotPoses

kielnino:
This is a PR for Issue #1083 

As discussed there, the definition of “is_fixed” in BundleShotPoses is probably incorrect, leading to unnecessary runs of the Ceres optimizer. Although the impact is rather minor, a correct definition is preferable.


kielnino:
Hi @YanNoun,

as your branch with the change for #1083 disappeared, I have set up a corresponding PR. Would you mind to merge it?

